PR #212, opened 2013-04-10
Made image class could use custom Drawer and Effects

AlloVince:
When using Imagine to create some special filters, I found that there is no way to change Imagine\Gd\Effects into self custom Class, so I think allow custom is necessary.

This pull request could make Imagine support custom Effects & Drawer

```
class Test extends Imagine\Gd\Effects {
    public function test(){}
}
$imagine = new Imagine\Gd\Imagine();
$image = $imagine->open(__DIR__ . '/Allo.png');
$image->effects('Test')->test();
```

### DIFF
--- a/lib/Imagine/Draw/DrawerInterface.php
+++ b/lib/Imagine/Draw/DrawerInterface.php
@@ -22,6 +22,14 @@ use Imagine\Image\PointInterface;
 interface DrawerInterface
 {
     /**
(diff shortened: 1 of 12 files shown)


mlocati:
IMHO #630 is a more general and cleaner approach...

mlocati:
Resolved in a more general way in #632

PR #242, opened 2013-08-08
Added new advanced filters

rejinka:
Again two new filters by me - i hope i did better with the style guidelines :-)

The two new filters are best descriped in the comments of the code. A test of BorderDetection is not necessary, i think, as it is only a nice shortcut to Neighborhood with nice matrices as options.


mlocati:
Updated in #644

PR #260, opened 2013-09-18
Allow slice image when copy source Image

ioalessio:
Defining an optional $size parameter in copy function i can create a new image with new size. Useful for image slice without resizing.


mlocati:
I agree with the comments of @stof and @romainneutron

PR #279, opened 2014-01-12
create BoxFactoryInterface to provide a easy way to change Box class imp...

psliwa:
Imagine library has `BoxInterface` interface but there are no way to change implementation that is used by Imagine. In my library that depends on Imagine I would like to change `Box` implementation, but there are no clean way to achieve this. So I have provided `BoxFactoryInterface` that can be injected to Imagine object. There are no any BC.


mlocati:
IMHO #630 is a more general and cleaner approach...

mlocati:
Resolved in a more general way in #632

PR #280, opened 2014-01-14
[FEATURE] Filter for making a round image.

chrootlogin:
Adds a new filter that creates a same sized empty copy of the image with transparent background. With the radius you set it calculates if a pixel is in- or outside the circle, if it's inside then it copies it to the new picture.


Richtermeister:
I would love to have a circular filter. I just wonder how this "copy pixels" approach stacks up against the approach of creating a circular mask and using it to mask the original image, in particular with regards to being able to anti-alias the edge of the circle.

Just curious, I haven't tried either approach myself yet.

mlocati:
I agree with @Richtermeister and @stof. Furthermore, we already have the great `applyMask` that works great, and it's much more flexible: you can use it to make an image round (or elliptical, or whatever). That's why I'm closing this.

PR #332, opened 2014-06-12
Added HHVM and PHP 5.6 to travis-ci

cebe:
run build against PHP 5.6 and hhvm but allow them to fail to just see how it works.

http://docs.travis-ci.com/user/build-configuration/#Rows-That-are-Allowed-To-Fail


burzum:
@cebe @romainneutron do you guys still need this PR? The develop branch already has 5.6 and 7.0 and just lacks HHVM support. Could you create a new branch / PR for HHVM?

cebe:
not really interested to work on this further. Too less time, sry.

PR #338, opened 2014-06-19
Removes counting layers to apply 'flatten' function

nlegoff:
Hello,

I've got an issue where I need to flatten an alpha channel against a white background of an image with a single layer.

See http://www.wizards-toolkit.org/discourse-server/viewtopic.php?f=3&t=20234#p80257 for more explanations.

One possible way to do that is to remove the check of counting layers to apply the function as if you apply the flatten option on a image with one layer it will not interfer at all.

We could also check if the image has an alpha channel (if it is possible with PHP Imagick) then authorize the appliance of the flatten function.


mlocati:
I do agree with @stof
I'm going to close this PR not because I'm against of it, but because too much time passed since it was created.
So, if you still have interest in it, feel absolutely free to complete it and reopen this PR.

Issue #162, opened 2012-11-14
final classes


ShurikAg:
Hi, 

We are using this library for our project and we came across an issue that blocks us from extending the functionality and in particular the fact that quite a few classes defined as final.

What is the actual reason for these classes to be final?

mlocati:
Me too I'd like to remove the final classes/methods.
We can follow semver if we have to break the backward compatibility.
So, I'll implement this before releasing the next version of Imagine (which should be 1.x) if there won't be any downvote from the other maintainers.

PR #339, opened 2014-06-22
Make Imagine more hackable, remove final keywords from classes and functions declaration

romainneutron:
This solves #162.

After thinking about it, I think we should open the interface so Imagine would be more hackable. I don't think classes should be extended, anyway, if some use case require it, let's give developers the ability to do so.

I propose to do it now as it would solve an issue toward version 1.0


mlocati:
I'm closing this because too many changes occurred. Let's continue the discussion in #162

PR #345, opened 2014-07-21
Add THUMBNAIL_UPSCALE setting

vlakoff:
This setting allows upscaling the source image if needed,
so that the result matches the requested size instead of being smaller.

### DIFF
--- a/lib/Imagine/Filter/Transformation.php
+++ b/lib/Imagine/Filter/Transformation.php
@@ -213,9 +213,9 @@ final class Transformation implements FilterInterface, ManipulatorInterface
     /**
      * {@inheritdoc}
      */
-    public function thumbnail(BoxInterface $size, $mode = ImageInterface::THUMBNAIL_INSET, $filter = ImageInterface::FILTER_UNDEFINED)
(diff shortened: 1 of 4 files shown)


stof:
@vlakoff note that I'm **not** the maintainer of this library. It is not me who need to review this to give it a merge.

vlakoff:
Indeed, but by bringing false alarms you may slow down the approval of the PR. We are at six months already. And yep I got a bit angry, because meanwhile I have to complexify *a lot* my code to have this feature...

stof:
@vlakoff my own reviews would allow to give it a better chance to be ready when @romainneutron reviews it.
Note that he is in vacation currently, so he won't review it in the coming days

tetronomis:
+1

vlakoff:
Please, don't [bump](https://en.wikipedia.org/wiki/Internet_forum#Bumping) this issue. That's only nine months. We are so close to a whole year. And I want to see if I die of old age before the patch makes it :)

mlocati:
I rebased this PR and created #645

PR #371, opened 2014-09-28
Typo fixed in GMagick\Imagine::create()

FractalizeR:
Typo fixed in GMagick\Imagine::create()


FractalizeR:
Well, this PR is about 4 years old. No wonder there were problems updating it ;) Glad we make it in the end. ;)

Issue #364, opened 2014-09-12
Imagine\Gd\Drawler::text() — problems with "fitted" text.


vlad-light:
Hello, I'm currently playing around with Imagine. So I decided to create an simple image-placeholder service... The user may pass a text-string which will be centered and added to the image. 
Consider following example:

``` php
//create required instances...
$Imagine = new \Imagine\Gd\Imagine();
$Palette = new \Imagine\Image\Palette\RGB();
//Background color of the upcomming image basend on settings passed by user...
$BackgroundColor = $Palette->color($settings['background_color'], 0);
//...some other code...
$TextColor = $Palette->color($settings['text_color'], 100);
//create an ImageBox for upcomming image based on given width and height
$ImageBox = new \Imagine\Image\Box($settings['width'], $settings['height']);
//get center X|Y coordinates for ImageBox
$ImageCenterPosition = new \Imagine\Image\Point\Center($ImageBox);
//create Text Object with font-family, font-size and color from settgins
$TextFont = new \Imagine\Gd\Font( $settings['font_file'] , $settings['font_size'],  $TextColor);
//create a Box (dimensions) based on font and given string...
$TextBox = $TextFont->box( $settings['text'] );
//get center X|Y coordinates for the box containing the text
$TextCenterPosition = new \Imagine\Image\Point\Center( $TextBox );
//calculate center X|Y coordanates from ImageBox center coordanates and FontBox center coordanates
$CenteredTextPosition = new \Imagine\Image\Point( $ImageCenterPosition->getX() - $TextCenterPosition->getX(), $ImageCenterPosition->getY() - $TextCenterPosition->getY());
//create an image with ImageBox dimensions and BackgroundColor
$_img_ = $Imagine->create($ImageBox, $BackgroundColor);
//now draw the text...
$_img_->draw()->text($settings['text'], $TextFont, $CenteredTextPosition);
```

this code works well (see list of cases and screens), untill I try to use `$width` param within a `draw()` method.
##### 1st case:

with following user settings creates — [Screenshot #1](https://cloud.githubusercontent.com/assets/277235/4243354/803523b0-3a0b-11e4-909d-ead87983c2ef.PNG) :ok_hand:

``` php
$settings['width'] = $settings['height'] = 600;
$settings['text'] = "some very long";
```

---
##### 2nd case:

[...]

PR #378, opened 2014-10-30
Fix for Issue #364.

vlad-light:
— added Font::wrapText() method which is a reflection of Drawer::wrapText()


vlad-light:
@stof  Ok, so what BC stands for ?

And what would be the right way ? Just copy the method-body to a right place whithout any reflection

stof:
backward compatibility

mlocati:
Implemented an alternative solution in #671

PR #395, opened 2014-11-12
Implementation of brightness effect (GD, Imagick, Gmagick)

lenybernard:


lenybernard:
I checked and you're right @stof, I wasn't testing all the drivers.
I wasn't able to take any time to finalize these tests but I've setup a vagrant and hope to be able to do this one night...

lib/Imagine/Gd/Effects.php
```diff
@@ -106,4 +106,16 @@ public function blur($sigma = 1)
 
         return $this;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function brightness($brightness)
+    {
+        if (false === imagefilter($this->resource, IMG_FILTER_BRIGHTNESS, $brightness)) {
```
mlocati: From the php manual it seems that brightness ranges from -255 to 255

mlocati:
Superseded bu #646

PR #401, opened 2014-12-17
Change the ImageMagick download URL

stof:
The ``download/`` folder only contains the archive for the latest release, and so the URL becomes invalid each time a new release is done. Using permanent URLs for the download is better


stof:
@romainneutron ping

harikt:
:( . not maintaining any more ?

stof:
@romainneutron ping

PR #410, opened 2015-02-25
Fix signature mismatch in IsImageEqual constraint

kitsunet:
The signature of \Imagine\Test\Constraint\IsImageEqual::evaluate does not match the one in the parent class \PHPUnit_Framework_Constraint. In PHP5.5 this leads to a runtime notice.
This change fixes the signature.


kitsunet:
The build fails because ImageMagick could not be downloaded...

romainneutron:
Thanks !

PR #418, opened 2015-03-16
Added trimImage option for Imagick

antonioperic:


romainneutron:
Hello and thanks for your work.

However, we can not integrate driver-only feature in imagine. If you want to create such feature, I recommend you to create your own service and apply them using Imagick\Image::getImagick method

igormukhingmailcom:
By the way there are driver-independent implementation of trim function.
See 4-th message at https://github.com/liip/LiipImagineBundle/issues/586
It has conditional for using `Imagick`'s built-in `trimImage` function for more performance.

antonioperic:
@igormukhingmailcom did you add this at the end?
I resolved my case with listener it works, but it will be nice to have trim filter inside

igormukhingmailcom:
@antonioperic I used that code (maybe with some small modifications) only on my project and not performed any PR yet.

antonioperic:
@igormukhingmailcom thnx for sharing it

igormukhingmailcom:
@antonioperic I thinking around how to test code right way. Because same `fuzz` may give different result for native `Imagick`'s and self-writed output. 

I think that should be some source image with linear gradient to test with. And both functions must remove same part of image during `trim`.

AlmogBaku:
Any update about adding @igormukhingmailcom PR? (well kinda PR)

antonioperic:
@AlmogBaku I don't think this option will be merged see comment at the begging. But from code you can see how to use it.

igormukhingmailcom:
@AlmogBaku Hi. That code example can't be added, because it designed to use as extension for LiipImagineBundle at AppBundle. So you can copy-paste it to your AppBundle to use.

PR #419, opened 2015-03-18
Ignore all test and documentation with the "export-ignore" attribute

Benoth:
Triming ~5Mb with composer --prefer-dist


romainneutron:
Thanks !

PR #421, opened 2015-03-22
support jpg extension

php-engineer:
It would be good to support also jpg extension or do you see any problems here?


burzum:
:+1:

stof:
This is already supported. See the implementation of ``normalizeFormat``

php-engineer:
You're right, sorry.

PR #429, opened 2015-05-06
Fix #428 : Fix ManipulatorInterface API docblocks

### LINKED ISSUE #428: Manipulator interface return type
Before doing any PR can we replace the ``@return ManipulatorInterface`` annotation with ``@return ImageInterface`` (it already implements ImageManipulator)  in the following file : https://github.com/avalanche123/Imagine/blob/develop/lib/Imagine/Image/ManipulatorInterface.php#L32 ?
Because it causes problems with Ide. For instance when we do : 

```php
$image->thumbnail(...)->getSize();
```

The ide cant find the method ``getSize``

romainneutron:


stof:
a bets solution would be to switch ManipulatorInterface to use ``@return static`` given that what it requires is returning an instance of the implementing class

romainneutron:
Oh well, this is indeed much much better @stof, thanks for pointing this

PR #446, opened 2015-07-10
Fix: Move files related to tests into appropriate directory

localheinz:
This PR

* [x] moves `Imagine\Test\ImagineTestCase` (and related) from `/lib` to `tests/`

:information_desk_person: Hey, these are tests, no!?


localheinz:
Thank you for looking into it, @romainneutron!

stof:
This makes these classes unavailable to people wanting to use them in their own testsuite. This is precisely why they were part of the library.

PR #447, opened 2015-07-10
Fix: Travis build configuration

localheinz:
This PR

* [x] moves updating of composer itself to `before_install` section
* [x] moves installation related steps to `install` section


localheinz:
@avalanche123 

Basically, just testing what's going on with the builds - it seems that all of them where `IMAGINE_DRIVER` is set to `imagick` are failing at the moment.

PR #433, opened 2015-05-06
Fix wrong array_merge usage

romainneutron:

PR #448, opened 2015-07-10
Fix: Download ImageMagick tar from appropriate location

localheinz:
This PR

* [x] uses the appropriate location from which to download the `ImageMagick-6.8.9-10` archive

:information_desk_person: Seems like the location has changed since the last PR (https://github.com/avalanche123/Imagine/pull/433) was merged.

See http://www.imagemagick.org/download/releases/.

/cc @romainneutron @stof


romainneutron:
Thanks !

PR #449, opened 2015-07-10
Getting applyMask to work

kasuparu:
I could not make GD\Image::applyMask() to work as described in Reflection example in http://www.slideshare.net/avalanche123/introduction-toimagine so I made some fixes.
1. It still supports only RGB palette for mask, but now accounts average between colors.
2. It does change image if its transparency is less than 0.5.

### DIFF
--- a/lib/Imagine/Gd/Image.php
+++ b/lib/Imagine/Gd/Image.php
@@ -376,7 +376,8 @@ final class Image extends AbstractImage
                 $position  = new Point($x, $y);
                 $color     = $this->getColorAt($position);
                 $maskColor = $mask->getColorAt($position);
-                $round     = (int) round(max($color->getAlpha(), (100 - $color->getAlpha()) * $maskColor->getRed() / 255));
+                $blackAmount = 1 - ($maskColor->getValue(ColorInterface::COLOR_RED) + $maskColor->getValue(ColorInterface::COLOR_GREEN) + $maskColor->getValue(ColorInterface::COLOR_BLUE)) / 3 / 255;
+                $round     = (int) round($color->getAlpha() * $blackAmount);
 
                 if (false === imagesetpixel($this->resource, $x, $y, $this->getColor($color->dissolve($round - $color->getAlpha())))) {
                     throw new RuntimeException('Apply mask operation failed');


spenibus:
Shouldn't the white parts of the mask stay visible and the black parts disappear ? While this seems to fix applyMask(), the mask seems to be applied backwards.

The following seems more appropriate:

    $whiteAmount = ($maskColor->getValue(ColorInterface::COLOR_RED) + $maskColor->getValue(ColorInterface::COLOR_GREEN) + $maskColor->getValue(ColorInterface::COLOR_BLUE)) / 3 / 255;
    $round       = (int) round($color->getAlpha() * $whiteAmount);

Using $blackAmount:
php-imagine-applymask-using-blackamount-20150731-1831-gmt

Using $whiteAmount:
php-imagine-applymask-using-whiteamount-20150731-1831-gmt

kasuparu:
"Reflection" example states that white is transparent, in other words, blacks stay and whites dissolve.

imagine-reflection

spenibus:
I guess I can only curse at the lack of standardisation regarding image masking. This really feels counter-intuitive.

axi:
Found this fix really usefull, should be merged

rvanlaak:
Does this resolve the bug that `applyMask` did not work before (see https://github.com/liip/LiipImagineBundle/issues/752)?

lib/Imagine/Gd/Image.php
```diff
@@ -376,7 +376,8 @@ public function applyMask(ImageInterface $mask)
                 $position  = new Point($x, $y);
                 $color     = $this->getColorAt($position);
                 $maskColor = $mask->getColorAt($position);
-                $round     = (int) round(max($color->getAlpha(), (100 - $color->getAlpha()) * $maskColor->getRed() / 255));
+                $blackAmount = 1 - ($maskColor->getValue(ColorInterface::COLOR_RED) + $maskColor->getValue(ColorInterface::COLOR_GREEN) + $maskColor->getValue(ColorInterface::COLOR_BLUE)) / 3 / 255;
```
romainneutron: This should be tested

PR #451, opened 2015-07-25
Fix open pdf with page parameter

JulienMalige:
Hi,

I'm currently using Imagine and I was confronted to a problem.

When I used Imagine to generate Images from PDF multi-pages files, Imagine generated me an image with another "image miniature" inside it. I wanted to generate a simple image and choose which page would be generate. When I was specifying a page (with .pdf[0]) like Imagick accept it, Imagine said to me that my file was not found (check path problem).

I asked a question on stack (I think it's more clear below) : 
http://stackoverflow.com/questions/31630229/imagine-imagick-cant-open-pdf-with-page-parameter

There is not many ressources or documentations about imagine working on PDF, so I made these modifications.

Hope this is not so dirty (my first contribution), but it seems work in my case.

Julien

### DIFF
--- a/lib/Imagine/Image/AbstractImagine.php
+++ b/lib/Imagine/Image/AbstractImagine.php
@@ -61,6 +61,13 @@ abstract class AbstractImagine implements ImagineInterface
             $path = (string) $path;
         }
 
+        $ext = pathinfo($path, PATHINFO_EXTENSION);
+
+        if(preg_match('/pdf\[[0-9]+\]/', $ext)){
+            $path = substr($path, 0, -strlen($ext));
+            $path = $path."pdf";
+        }
+
         $handle = @fopen($path, 'r');
 
         if (false === $handle) {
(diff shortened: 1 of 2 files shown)


srdjanmarjanovic:
+1 for this! As I can see, the build is not passing due to composer errors... This is realy a big issue for my team - hence this bump.

JulienMalige:
@srdjanmarjanovic it was my first try of contribution. I will work on this before Sunday !

JulienMalige:
@srdjanmarjanovic in fact this project not seems to be still maintained...

mlocati:
The purpose of this pull request is to change the $path parameter, to replace extensions like `.pdf[0]` with `.pdf`, right?
If so, how could people load a file that actually ends with  `.pdf[0]`?

JulienMalige:
@mlocati sorry but it was an old PR. I think the need was outdated, no problem for closed it :)

PR #453, opened 2015-07-28
Unsilence GD Library errors in Imagine\Gd\open()

jasonquick:
I think silencing errors here is a bad idea because it obliterates potentially useful/vital information as to why the image could not be opened. With this error silenced a developer can only guess. In my case on a fresh PHP install I had forgotten to enable JPEG support. I spent several minutes looking for a remote URL issue (as in my case the image is opened from a remote URL) before I figured out the GD Library error was being silenced.

Is there a particular rationale for silencing these messages?

### DIFF
--- a/lib/Imagine/Gd/Imagine.php
+++ b/lib/Imagine/Gd/Imagine.php
@@ -90,7 +90,7 @@ final class Imagine extends AbstractImagine
             throw new RuntimeException(sprintf('Failed to open file %s', $path));
         }
 
-        $resource = @imagecreatefromstring($data);
+        $resource = imagecreatefromstring($data);
 
         if (!is_resource($resource)) {
             throw new RuntimeException(sprintf('Unable to open image %s', $path));


jasonquick:
Okay, okay, I broke it. I'm working on an updated test...

stof:
The rational is that the library says it will throw a RuntimeException on errors, not throw a warning. If it throws a warning first and you have an error handler turning them into exceptions (most frameworks do), you will get the wrong exception, breaking the code handling Imagine errors (as it won't be reported as an Imagine error).

The right way would be to retrieved the error message of the GD warning and include it in the exception message.

stof:
or register a custom error handler around the opening call. This might be an option

jasonquick:
That's the direction I was headed, but without taking some time to understand how Imagine expects errors to be handled, and if there's already custom error_handlers elsewhere I was a little reluctant to put too much effort into it. I'm not unwilling to tackle it, but it strikes be as being trivial for someone more plugged-in to the code base.

PR #462, opened 2015-09-20
Provide detected version in exception message

romainneutron:


lib/Imagine/Gd/Imagine.php
```diff
@@ -178,7 +178,7 @@ private function loadGdInfo()
     private function requireGdVersion($version)
     {
         if (version_compare(GD_VERSION, $version, '<')) {
-            throw new RuntimeException(sprintf('GD2 version %s or higher is required', $version));
+            throw new RuntimeException(sprintf('GD2 version %s or higher is required, %s provided', $version, GD_VERSION));
```
stof: I would rather say ``detected`` than ``provided``

PR #464, opened 2015-10-07
Feature/smart crop

guillaumeogier:
This pull request include two smart crop to improve the original crop proposed by imagine. 
One is a crop by entropy and second is a crop by balanced.

thanks

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "imagine/imagine",
+    "name": "inneair/imagine",
     "description": "Image processing for PHP 5.3",
     "keywords": [
         "image manipulation",
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "inneair/imagine",
+    "name": "imagine/imagine",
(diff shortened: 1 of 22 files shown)


guillaumeogier:
If you want to see how it work or just see the difference between theses crops and the classical you can go to https://github.com/guillaumeogier/smart-crop, and watch the doc or the result of cropping

lib/Imagine/Effects/EffectsInterface.php
```diff
@@ -77,4 +77,15 @@ public function sharpen();
      * @throws RuntimeException
      */
     public function blur($sigma);
+
```
Koc: file permissions here and below whould be reverted

Koc:
@guillaumeogier can you also update documentation?

burzum:
@Koc can you do and do another PR? I'm about to go to bed. I've missed the permissions. :( My bad.

Koc:
I know nothing about this feature, so it would more logical that contributor will add docs about new things that he introduced

PR #468, opened 2015-11-13
Workaround an issue where gmagick segfaults.

meckhardt:
This approach works around an anoying segfault when gmagick instances are destroyed.

It might be that memory is messed a bit, but better then segfaults.


meckhardt:
I have seen no issues regarding memory consumption in my production environment, which uses Imagine to convert images on the fly (3k req / minute).

mlocati:
Closing because of lack of feedback.

PR #469, opened 2015-11-17
Enable past with alpha transparency

drmovi:
set alpha transparency for overlay image

### DIFF
--- a/lib/Imagine/Gd/Image.php
+++ b/lib/Imagine/Gd/Image.php
@@ -131,7 +131,7 @@ final class Image extends AbstractImage
      *
      * @return ImageInterface
      */
-    final public function paste(ImageInterface $image, PointInterface $start)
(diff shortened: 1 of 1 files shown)


stof:
having a different signature than the interface does not make sense. It defeats the point of being an abstraction library

drmovi:
This is suggestion to add transparency feature to past function

stof:
@drmovi but then, it must be added in the interface and implemented for all libraries.

And I'm not sure changing the existing method is a good idea. It might have unwanted side effects

drmovi:
3 libraries , i think that is not a big deal

burzum:
@drmovi any plans to finish this PR? I agree with @stof, it doesn't make sens to merge this without his proposed changes.

mlocati:
Implemented in #639

PR #475, opened 2015-12-09
Reverse deprecated version logic.

### LINKED ISSUE #474: PHP7
on PHP7 & [phpseven-imagick](https://github.com/mkoppanen/imagick/tree/phpseven) 
I get 
```
ErrorException: Imagick::clone method is deprecated and it's use should be avoided in
vendor/imagine/imagine/lib/Imagine/Imagick/Image.php:82
```

Because:
```
$ php7 -r 'var_dump(phpversion("imagick"));'
string(17) "@PACKAGE_VERSION@"

$ php7 -r 'var_dump(version_compare(phpversion("imagick"), "3.1.0b1", ">="));'
bool(false)

```

Danack:
Imagick does not have a version contained inside it, unless it was packaged by PECL. For versions compiled from source, the version is not a semver number, and so can't be compared sanely. This change makes the code assume that Imagick is being compiled from source, then it is from post 3.1.0b1, and not some code that was released on in February 2011.

Fixes #474

### DIFF
--- a/lib/Imagine/Imagick/Image.php
+++ b/lib/Imagine/Imagick/Image.php
@@ -105,10 +105,10 @@ final class Image extends AbstractImage
     public function copy()
     {
         try {
-            if (version_compare(phpversion("imagick"), "3.1.0b1", ">=") || defined("HHVM_VERSION")) {
-                $clone = clone $this->imagick;
-            } else {
+            if (version_compare(phpversion("imagick"), "3.1.0b1", "<") && !defined("HHVM_VERSION")) {
                 $clone = $this->imagick->clone();
+            } else {
+                $clone = clone $this->imagick;
             }
         } catch (\ImagickException $e) {
             throw new RuntimeException('Copy operation failed', $e->getCode(), $e);


githubjeka:
Fix isn't for me :(

image

var_dump(phpversion("imagick"))
string(17) "@PACKAGE_VERSION@"

githubjeka:
Why not use the `clone $this->imagick;` always ?

Danack:
Because PHP...

```
var_dump(version_compare('bananas', "3.1.0b1", "<"));
```
result is true ..... bool(true).

So yeah, the implementation of this PR is bogus, but the idea is there. And there maybe better ways of doing it.

lib/Imagine/Imagick/Image.php
```diff
@@ -105,10 +105,10 @@ public function getImagick()
     public function copy()
     {
         try {
-            if (version_compare(phpversion("imagick"), "3.1.0b1", ">=") || defined("HHVM_VERSION")) {
-                $clone = clone $this->imagick;
-            } else {
+            if (version_compare(phpversion("imagick"), "3.1.0b1", "<") && !defined("HHVM_VERSION")) {
```
zamaliphe: phpversion("imagick") still return @PACKAGE_VERSION@ when compiled on php7

mlocati:
Superseded by #628

PR #479, opened 2016-02-04
Avoid using 2.3x the image size worth of memory when no memory is enough

Seldaek:
exif_read_data can read from the filesystem just fine, so if we get an image path, doing file_get_contents() will load up the file in memory, and then passing it to doReadData will base64_encode it which typically adds 1/3rd of the binary size to the output, so you add 1.3x the image size into the memory. 

This is all fine with thumbnail sized images but when dealing with large files you very easily and unnecessarily blow up the memory limit.

There is no 0.6 branch but this applies cleanly on 0.6.3 as well, in case you'd like to do a new release :)

### DIFF
--- a/lib/Imagine/Image/Metadata/ExifMetadataReader.php
+++ b/lib/Imagine/Image/Metadata/ExifMetadataReader.php
@@ -36,6 +36,14 @@ class ExifMetadataReader extends AbstractMetadataReader
      */
     protected function extractFromFile($file)
     {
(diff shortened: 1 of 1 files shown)


Seldaek:
@romainneutron just so I know are you still maintaining this or is it completely abandoned? :)

romainneutron:
@Seldaek Sorry for the late reply. Thanks for this !

PR #437, opened 2015-05-27
Ensure all layers of animated GIFs are preserved when cropped

armatronic:
Currently Imagick only support.

PR #348, opened 2014-07-29
support resize of animated gifs with imagick

scuben:
support of resizing images with multiple layers with imagick.

feedback very welcome!

PR #481, opened 2016-02-23
Preserve gif animations

trsteel88:
This is a combination of #348 and #437.


dupuchba:
where are we with this one ? @romainneutron

trsteel88:
I'm actually a little worried about this one. 

If someone uploads an image with 1,000 frames, that is 1000 images that need to be processed I think? A 10000 by 10000 pixel image could cause quite a bit of load in that instance.

dupuchba:
@trsteel88 agree with you but as a a user of Imagine, it's upon to you to make the choice to allow GIF or not.
Personally, I use this special code at a client's (a big french application) and it works perfectly.

What's left to let it merged ?

trsteel88:
I think it's good to go. 

Would probably be a good idea to add a frame limit. So if that was 10 and there was 1000 frames it would take every 100th frame.

dupuchba:
@trsteel88 why not, as long as we manage to document it well! Maybe a configurable frame limit ?

burzum:
@trsteel88 can you add the frame limit option? :)

lib/Imagine/Imagick/Image.php
```diff
@@ -127,15 +127,24 @@ public function crop(PointInterface $start, BoxInterface $size)
...
+                $this->imagick = $this->imagick->deconstructImages();
+            } else {
+                $this->imagick->cropImage($size->getWidth(), $size->getHeight(), $start->getX(), $start->getY());
+                // Reset canvas for gif format
+                $this->imagick->setImagePage(0, 0, 0, 0);
+            }
         } catch (\ImagickException $e) {
             throw new RuntimeException('Crop operation failed', $e->getCode(), $e);
         }
-
```
romainneutron: Wrong CS

hjanuschka:
@burzum  any chances on getting a new release?

Issue #466, opened 2015-10-26
Incorrect rgb color blend algorithm 


dmolineus:
The color blend algorithm doesn't work with all cases. For instance doing a blend from yellow to red will flip the colors, so that a blend from red to yellow is drawn. 

Correct implementation would be (based on http://stackoverflow.com/a/27553):

``` php
        return $this->color(
            array(
                (int) $color2->getRed() * $amount + $color1->getRed() * (1 -$amount),
                (int) $color2->getGreen() * $amount + $color1->getGreen() * (1 -$amount),
                (int) $color2->getBlue() * $amount + $color1->getBlue() * (1 -$amount),
            ),
            (int) min(100, min($color1->getAlpha(), $color2->getAlpha()) + round(abs($color2->getAlpha() - $color1->getAlpha()) * $amount))
        );
```

romainneutron:
Could you send a PR?

dmolineus:
Done with #488.

mlocati:
Fixed in #649

PR #488, opened 2016-03-11
Hotfix color gradients.

dmolineus:
This pull request provides a fix for the broken color gradients calculation mentioned in #466.


mlocati:
Rebased in #649

PR #490, opened 2016-03-14
If box is a square, providing width is enough

gabrielkaputa:
This pull request provides an option to provide only $width when the Box is intended to be a square

### DIFF
--- a/lib/Imagine/Image/Box.php
+++ b/lib/Imagine/Image/Box.php
@@ -36,8 +36,12 @@ final class Box implements BoxInterface
      *
      * @throws InvalidArgumentException
      */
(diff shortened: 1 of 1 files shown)


lib/Imagine/Image/Box.php
```diff
@@ -36,8 +36,12 @@
      *
      * @throws InvalidArgumentException
      */
-    public function __construct($width, $height)
+    public function __construct($width, $height = null)
```
romainneutron: This behavior is not documented.

Moreover, I'd have used a static shortcut like `Box::square($size)` instead of an optional `height`

PR #495, opened 2016-05-19
Remove unnecessary PDF from autogenerated ZIP archives

mlocati:
What about removing this (big) PDF from autogenerated ZIP archives?
I don't think it's useful for production machines, developers that need this may still do a `git clone`.

Are the other PDF files needed for legal purposes or can we remove them too?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,6 +5,7 @@
 /docs               export-ignore
 /tests              export-ignore
 /lib/Imagine/Test   export-ignore
+/lib/Imagine/resources/Adobe/Profile[[:space:]]Information.pdf export-ignore
 /.gitattributes     export-ignore
 /.gitignore         export-ignore
 /.travis.yml        export-ignore


burzum:
@avalanche123 is this included for legal reasons? I would merge it because I agree with @mlocati but I'm not sure about possible legal reasons.

.gitattributes
```diff
@@ -5,6 +5,7 @@
 /docs               export-ignore
 /tests              export-ignore
 /lib/Imagine/Test   export-ignore
+/resources/Adobe/Profile Information.pdf export-ignore
```
romainneutron: path is `/lib/Imagine/resources/Adobe/Profile`

mlocati: @romainneutron You're right: I updated this PR

mlocati: @romainneutron I had also to use `[[:space:]]` instead of the space in the file name

mlocati: Done (one year ago)

Issue #496, opened 2016-05-23
New Imagine7 constants


Taluu:
Hey,

It seems some constants are just now gone. cf mkoppanen/imagick#151

I guess some other constants may have changed, but which ones...

alpharder:
👍

burzum:
Closing this because #500 should fix this. Please respond if #500 didn't fix it and I'll reopen the ticket.

PR #500, opened 2016-06-15
Fix imagine7 constants

joschi127:
Not so nice but working fix for https://github.com/avalanche123/Imagine/issues/496

Does anyone else know a better solution?

The constants seem to be gone. I could not find new constants to use.

### DIFF
--- a/lib/Imagine/Imagick/Image.php
+++ b/lib/Imagine/Imagick/Image.php
@@ -809,9 +809,15 @@ final class Image extends AbstractImage
     {
         static $typeMapping = array(
             // We use Matte variants to preserve alpha
-            PaletteInterface::PALETTE_CMYK      => \Imagick::IMGTYPE_TRUECOLORMATTE,
-            PaletteInterface::PALETTE_RGB       => \Imagick::IMGTYPE_TRUECOLORMATTE,
-            PaletteInterface::PALETTE_GRAYSCALE => \Imagick::IMGTYPE_GRAYSCALEMATTE,
+            //
+            // (the constants \Imagick::IMGTYPE_TRUECOLORMATTE and \Imagick::IMGTYPE_GRAYSCALEMATTE do not exist anymore in Imagick 7,
+            // to fix this the former values are hard coded here, the documentation under http://php.net/manual/en/imagick.settype.php
+            // doesn't tell us which constants to use and the alternative constants listed under
+            // https://pecl.php.net/package/imagick/3.4.3RC1 do not exist either, so we found no other way to fix it as to hard code
+            // the values here)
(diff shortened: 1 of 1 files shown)


andris-sevcenko:
You can find the correct constants in the Imagick PECL [changelog](https://pecl.php.net/package-info.php?package=imagick&version=3.4.3RC1)

takobell:
Looks like it's not just these enums that have changed, there are several others listed in the changelog that Imagine is using as well https://github.com/mkoppanen/imagick/issues/151#issuecomment-225014365 and https://pecl.php.net/package-info.php?package=imagick&version=3.4.3RC1

joschi127:
thanks :+1:

robfrawley:
@burzum We should add a `@todo` or other note in the comments, as hard coding these values *will* backfire at some point once they stop being supported. They were removed for a reason, so we should either figure out what constants replace them, or otherwise what functionality replaced them. Otherwise, this will come back to bite us at some version in the future.

Not against this merge or fix, but we need to document in a way that can be easily found via a regular expression or otherwise so we can locate places in the code that are possibly going to cause issues down the road.

lib/Imagine/Imagick/Image.php
```diff
@@ -807,11 +807,17 @@ private function getMimeType($format)
      */
     private function setColorspace(PaletteInterface $palette)
     {
-        static $typeMapping = array(
```
romainneutron: why removing the static definition?

romainneutron: It removes memory optimization on PHP5

romainneutron: Fixed in 816a34d

robfrawley:
This should probably be handled using the proper constants:

- `IMGTYPE_TRUECOLORMATTE` -> `IMGTYPE_TRUECOLORALPHA`
- `IMGTYPE_GRAYSCALEMATTE` -> `IMGTYPE_GRAYSCALEALPHA`

See https://github.com/Intervention/image/pull/679

joschi127:
Also see my source code comment starting in line 813

https://github.com/joschi127/Imagine/blob/bbb008c076e167e414153a740c853433fd636754/lib/Imagine/Imagick/Image.php#L813

I think I have already tried to use the correct constants but these were not available as well. I do not know why. Maybe only with certain, maybe broken, versions of the Imagick extension or maybe I did something wrong. If you want we can try again and I can also test again on the server environment which was affected when implementing this fix.

joschi127:
Also found these comments here:

So it sounds like  is broken with ImageMagick 7.0.4-6 and imagick 3.4.2_7 and fixed with imagick >= 3.4.3.

https://github.com/Intervention/image/pull/679#issuecomment-277670559

So if imagick >= 3.4.3 is now available on all stable distributions we can remove the work around with hard coded constants again?

PR #505, opened 2016-08-03
png compression method changed in imagemagick 6.8.8

chregu:
Looks like starting with imagemagick 6.8.8, you have to use `setCompressionQuality` instead of `setImageCompressionQuality`.

### DIFF
--- a/lib/Imagine/Imagick/Image.php
+++ b/lib/Imagine/Imagick/Image.php
@@ -696,8 +696,15 @@ final class Image extends AbstractImage
             } else {
                 $compression += 5;
             }
-
-            $image->setImageCompressionQuality($compression);
+            $v = \Imagick::getVersion();
+            preg_match('/ImageMagick ([0-9]+\.[0-9]+\.[0-9]+)/', $v['versionString'], $v);
+            if (version_compare($v[1], '6.8.7') < 0 ) {
+                //Use this for ImageMagick releases before 6.8.7-5
+                $image->setImageCompressionQuality($compression);
+            } else {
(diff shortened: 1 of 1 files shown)


lib/Imagine/Imagick/Image.php
```diff
@@ -696,8 +696,15 @@ private function applyImageOptions(\Imagick $image, array $options, $path)
...
             }
-
-            $image->setImageCompressionQuality($compression);
+            $v = \Imagick::getVersion();
+            preg_match('/ImageMagick ([0-9]+\.[0-9]+\.[0-9]+)/', $v['versionString'], $v);
+            if (version_compare($v[1], '6.8.7') < 0 ) {
+                //Use this for ImageMagick releases before 6.8.7-5
+                $image->setImageCompressionQuality($compression);
+            } else {
+            //Use this for ImageMagick releases after 6.8.7-5
```
burzum: Can we get the indentation right here? Also shouldn't there be a space after the //?

lib/Imagine/Imagick/Image.php
```diff
@@ -696,8 +696,15 @@ private function applyImageOptions(\Imagick $image, array $options, $path)
             } else {
                 $compression += 5;
             }
-
-            $image->setImageCompressionQuality($compression);
+            $v = \Imagick::getVersion();
+            preg_match('/ImageMagick ([0-9]+\.[0-9]+\.[0-9]+)/', $v['versionString'], $v);
```
romainneutron: We should share the code with Imagick\Imagick\Imagine#L173

PR #520, opened 2016-11-22
fix for png compression in imagick adapter

lemorragia:
Hi, 
if i understood correctly, there was a little difference in setting the png compression between the gd and the imagick adapter. 
In the imagick adapter the png_compression_level and png_compression_filter were never set by the "quality" parameter. The gd adapter had a different behaviour.
I moved the code from the gd adapter to the AbstractImage class which both extends to ensure that the parameters are correctly set in both cases. It's a change which doesn't break BC in the gd adapter, and the inherent code for handling the compression in the imagick adapter was already there.

### DIFF
--- a/lib/Imagine/Gd/Image.php
+++ b/lib/Imagine/Gd/Image.php
@@ -539,14 +539,6 @@ final class Image extends AbstractImage
         $save = 'image'.$format;
         $args = array(&$this->resource, $filename);
 
-        // Preserve BC until version 1.0
-        if (isset($options['quality']) && !isset($options['png_compression_level'])) {
-            $options['png_compression_level'] = round((100 - $options['quality']) * 9 / 100);
-        }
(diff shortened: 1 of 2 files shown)


burzum:
@lemorragia can we get a test for this please?

@avalanche123 I can't re-run the tests, would you mind to run this one again? Looks like it could not fetch a dependency and failed for that reason.

lemorragia:
@burzum i'm not quite sure how to test this given that is a pre-existing behaviour. I'm reading the tests and apparently the compression (both gd and imagick) is not currently tested...ideas to test if the object is correctly set with the selected compression? Maybe Imagick::getImageCompressionQuality will do the trick for Imagick, but i have no idea for GD

mlocati:
I think that the quality option is a good fallback value for any saved format.
That's why in #619 I adopted the opposite aporoach of this PR: use quality if the format specific compression level isn't available.

PR #525, opened 2017-02-02
Fix Imagick "rotate" page alignment

thePanz:
Some transformations including Imagick ::rotateImage() may change "image page" -- working area inside the image you work on.
Be careful with future modifications afterwards because the image page would be different from new sizes of the image.
For example, if you do Imagic::cropImage() after rotation, you need to set image page properly, otherwise your crop would be performed relating to wrong coordinates (depending on rotation angle, resulting image size may vary).

See http://php.net/manual/en/imagick.rotateimage.php#119455


romainneutron:
it would have been great to add a unit test

PR #526, opened 2017-02-02
Fix Imagick 'rotate': reset canvas size after operation

thePanz:
See: http://php.net/manual/en/imagick.rotateimage.php#119455

> Some transformations including Imagick ::rotateImage() may change "image page" -- working area inside the image you work on.
Be careful with future modifications afterwards because the image page would be different from new sizes of the image.
For example, if you do Imagic::cropImage() after rotation, you need to set image page properly, otherwise your crop would be performed relating to wrong coordinates (depending on rotation angle, resulting image size may vary).

### DIFF
--- a/lib/Imagine/Imagick/Image.php
+++ b/lib/Imagine/Imagick/Image.php
@@ -337,6 +337,8 @@ final class Image extends AbstractImage
             $pixel = $this->getColor($color);
 
             $this->imagick->rotateimage($pixel, $angle);
+            // Reset the image page: next operations will use the correct canvas size
+            $this->imagick->setImagePage(0, 0, 0, 0);
 
             $pixel->clear();
             $pixel->destroy();


burzum:
Won't this change the behavior of the current implementation and lead to unexpected results?

thePanz:
@burzum if the "unexpected" result is to properly render an image, yes :)

I agree with you that the change is quite important, but the current implementation is broken: try the following:
- Create an image 200x200
- Rotate the image
- Crop the image to 500x500

The whole image should be placed in the center (as the bigger cropping size) and rotated. But as for the current implementation the corners are cut in an expected way. The comment in the PHP page is quite clear about the issue here

burzum:
@thePanz yes, I understand that. But what happens if people have already worked around that? :)

I think this can be merged and a new version must be tagged  then. I'm trying to fix the tests right now.

romainneutron:
A unit test should have been welcome

thePanz:
@romainneutron: I can provide a test for that, I've coded it for a service that internally uses Imagine

Issue #521, opened 2016-12-07
Is this project dead?


burzum:
There is a huge amount of good PRs and an even bigger amount of issues that don't get any attention.

Why? Is the project dead?

I really consider switching to https://github.com/Intervention/image because of the lack of support for this lib. :(

scoolen:
ping @avalanche123 @romainneutron

avalanche123:
would you like to be the maintainer?

scoolen:
Not sure if directed at me but:
No thanks, not at this time.

burzum:
@avalanche123 well, not really, I'm involved a lot with CakePHP and my own plugins for it (there is an imagine plugin). But if you like you can give me permissions to merge PRs. There is a ton of them waiting for approval and I think there are a few good ones present.

Also I would like to set up Scrutinizer CI for the project besides Travis to get code coverage and code analysis and phpcs checks. Are you using the PSR standard?

avalanche123:
@burzum sure, I'll add you to collaborators

burzum:
Thank you! You don't have the time to maintain this project any more? :(

I don't mind helping out sometimes but it would be also great to have another person reviewing my own changes if I decide to do some.

Also how is your opinion on dropping support for php up to 5.5?

avalanche123:
Yes, let's drop it. I don't mind reviewing your pull requests, you might have to ping me explicitly. Yes, I'd appreciate any help with maintenance that you can give.

burzum:
OK, I'll let you know. So far I haven't had a need to fix anything inside the lib. :) Great work you've done so far!

avalanche123:
Thanks!

thePanz:
Thank you @burzum to take this project over! :)
Could you please tag a new release? :)

burzum:
@thePanz I'm **not** taking it over, just trying to help out with the PR's and issues.

Not sure if it's OK to tag a new release, I prefer if the maintainer would do that. I don't want to make such decisions without the project lead.

thePanz:
Ok, clear! :) Btw: seems that most of the tests are now failing due to a version mismatch :(
I guess this repo needs some more love.. I'll try to find some time to help you out here, is it Ok for you?

burzum:
Sure! Just contact @avalanche123 and ask him if you can get collaborator rights as well.

avalanche123:
@thePanz I've added you as a collaborator

robfrawley:
[...]

PR #528, opened 2017-02-02
Fixing tests

burzum:


robfrawley:
Re: https://github.com/avalanche123/Imagine/issues/521#issuecomment-277558880

Just pushed a commit to #528 with the changes need to use `phpunit` version `5.3`. Prior, there was no `script` section, which means Travis will use the "global" (system installed) version of `phpunit`. Installing it using "global" `composer`, as you were trying, does so in the "user" scope (which would still need a special runtime configuration to be called). I removed that, as the `composer.json` already specifies the version requirement, thus it will be installed in the "project" scope. At this point, it can be called by the addition I made to `composer.json`:

```yml
script:
  - ./bin/phpunit -vvv
```

Still failing, but it is using the version of `phpunit` desired.

robfrawley:
Also, I removed the `composer self-update` command, as Travis already does this during its bootstrapping process. See [line 138](https://travis-ci.org/avalanche123/Imagine/jobs/198689121#L138) from a recent build.

robfrawley:
@burzum The `hirak/prestissimo` package need to be added globally for it to have any affect; moreover some configuration needs to be in place. I recently added it to the Travis build for `liip/imagine-bundle` and you can reference my pull request there for inspiration: liip/LiipImagineBundle#864.

romainneutron:
Superseeded by #530

PR #532, opened 2017-02-09
Revert "Getting applyMask to work"

romainneutron:
Please reopen with a unit test


rvanlaak:
Would be great to have this fixed, as now the crop does not seem to work. @romainneutron what should the unit test cover?

PR #533, opened 2017-02-09
Revert "Feature/smart crop"

romainneutron:
This PR should be discussed. It should probably be placed in a "community library" extension outside imagine


romainneutron:
Please elaborate a brung constructive discussion instead of reacting with "thumbs up" or "thumbs down". It does not bring value, just make me feel bad, which is not a nice feeling when you spend time providing time to others for free (aka opensource).

As I already mentioned in https://github.com/avalanche123/Imagine/issues/521#issuecomment-278639899, refusing a contribution is not an attack. It's defining the perimeter definition of what will be maintained and guarantee by Imagine (it's a matter of responsability).

As we want to provide an extensible library, it could be very nice to improve extensibility instead of adding features to the core.

VincentClair:
@romainneutron Sorry, I don't want to make you feel bad. And I don't take the revert as an attack.

I fully understand the need to border the perimeter of the library.
My main idea behind this was to get a library that don't just manipulate images, but add value on based functionalities. It's only MHO: for example, personally, in all projets i have made, i don't need a drawing API, but the masking functionality is enough.

Many of my clients asked me for a better crop. I have put some logic in a back-office, so they could crop the image them-self, or add a focus point for a responsive image. But a smart crop is more interesting for me if it could be done natively and simplify UX. And i prefer to install one library than many plugins. As I said, it's only my needs.

PR #543, opened 2017-02-13
Getting applyMask to work

romainneutron:
From original PR #449


mlocati:
@romainneutron is this meant to be merged?

mlocati:
This shouldn't be need. IMHO the original issue was caused by the user not calling `mask()` on the mask

Commandor05:
> This shouldn't be need. IMHO the original issue was caused by the user not calling `mask()` on the mask

Could you clarify what you mean? Because i have same problem with applyMask after rollback.

PR #546, opened 2017-02-13
Fix Imagick 'rotate': reset canvas size after operation

romainneutron:
See #526


mlocati:
I see no change in this pull request...

PR #549, opened 2017-03-04
Don't unnecessarily copy image in ->thumbnail()

garygreen:
Not sure why it's copying the image again when you generating a thumbnail? Seems like an unnecessary operation....


romainneutron:
This code should be outside the Image itself. The copy here is to avoid modifying the object itself and be able to create multiple thumbnails from a single original. Otherwise it would be impossible

PR #551, opened 2017-03-08
Fix missing arg on last version of Imagick

jdespatis:
Impossible to colorize an image with the Imagick driver

Problem explained at:
mkoppanen/imagick#171

### DIFF
--- a/lib/Imagine/Imagick/Effects.php
+++ b/lib/Imagine/Imagick/Effects.php
@@ -81,7 +81,7 @@ class Effects implements EffectsInterface
         }
 
         try {
-            $this->imagick->colorizeImage((string) $color, new \ImagickPixel(sprintf('rgba(%d, %d, %d, 1)', $color->getRed(), $color->getGreen(), $color->getBlue())));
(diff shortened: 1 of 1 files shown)


lib/Imagine/Imagick/Effects.php
```diff
@@ -81,7 +81,7 @@ public function colorize(ColorInterface $color)
         }
 
         try {
-            $this->imagick->colorizeImage((string) $color, new \ImagickPixel(sprintf('rgba(%d, %d, %d, 1)', $color->getRed(), $color->getGreen(), $color->getBlue())));
+            $this->imagick->colorizeImage((string) $color, new \ImagickPixel(sprintf('rgba(%d, %d, %d, 1)', $color->getRed(), $color->getGreen(), $color->getBlue())), true);
```
romainneutron: Should be false (see https://github.com/mkoppanen/imagick/blob/63e9f6bc5184d601298aaa54aebca0ed7cf0d12b/imagick_class.c#L8930-L8932)

mlocati:
@jdespatis could you update this PR accordingly to what @romainneutron suggests?

mlocati:
The value of the `legacy` argument is by default set to `false` (and [we want it to be false](https://github.com/mkoppanen/imagick/blob/63e9f6bc5184d601298aaa54aebca0ed7cf0d12b/imagick_class.c#L8930-L8932)), so this change is unnecessary.

Issue #515, opened 2016-10-10
Using supplied ExifMetadataReader fails (when not ignoring @suppressed errors with custom error handler)


twicejr:
In my framework, I do not allow the use of @ in code; or at least, the errors are still passed.
I get the error: "File not supported" by exif_read_data.

Why not use a try/catch statement to prevent this 'bug' happening?

```
     //Error handler for handling exceptions.
    set_error_handler(function($severity, $message, $file, $line)
    {
        // error suppressed with @ so still report it!!!!!!!!
        if (error_reporting() === 0) {
            throw new ErrorException($message, $severity, $severity, $file, $line);
        } 

        if(!(error_reporting() & $severity))
        {
            return;
        }
        throw new ErrorException($message, $severity, $severity, $file, $line);
    });

    try
    {
        if (false === $exifData = exif_read_data($path, null, true, false)) {
            return array();
        }
    }
    catch(Exception $e)
    {
        return array();
    }

    restore_error_handler();
```

Instead of

```
   if (false === $exifData = @exif_read_data($path, null, true, false)) {
            return array();
   }
```

matej-prazek:
I want to suggest more clear way to fix this bug because `@` is awful (set and restore error handler is not good too).

`exif_read_data` can accept only JPG and TIFF files (http://php.net/manual/en/function.exif-read-data.php)

```text
exif_read_data — Reads the EXIF headers from JPEG or TIFF:
```

That's why:

```
$imageType = exif_imagetype($path);
$exifData = false;
if (in_array($imageType, array(IMAGETYPE_JPEG, IMAGETYPE_TIFF_II, IMAGETYPE_TIFF_MM))) {
     $exifData = exif_read_data($path, null, true, false);
}
if (false === $exifData) {
     return array();
}
```

esprit-dev:
I've the same issue. Is it possible to fix it ? That makes huge log files in some case where warning errors are recorded (mass processing, etc.)

twicejr:
You have to check if it is jpg , otherwise do not check the metadata..
You can also make it catchable by making your own error_handler catch every
error, even suppressed ones.

Op 6 jan. 2017 4:33 p.m. schreef "Frédéric TISSOT" <notifications@github.com

twicejr:
//Error handler for handling exceptions.
        set_error_handler(function($severity, $message, $file, $line)
        {
            // error suppressed with @ so still report it!!!!!!!!
            if (error_reporting() === 0) {
[...]

PR #554, opened 2017-03-26
Fix exif_read_data warning 'File not supported'

TumbleweedBY:
Problem with set_error_handler [https://github.com/avalanche123/Imagine/issues/515](https://github.com/avalanche123/Imagine/issues/515).


robfrawley:
What error reporting level are you using in production that ignores the "shut-up" operator?

mlocati:
I'd prefer to temporarily set an error handler, for two reasons:
1. We don't have to track changes in the exif library if it starts suporting new formats
2. We can intercept other warnings possibly raised by `exif_read_data`

mlocati:
This PR has been replaced by #620

PR #555, opened 2017-04-10
[Docs] added "Create Thumbnails" section

mktcode:
Using the resize method to resize an image without distortion is not that intuitive. Using the thumbnail method is much easier but it is not explained in this document.


romainneutron:
Please document precisely what's the difference with the `resize` method. IIRC it resizes and applies other operations

PR #558, opened 2017-05-02
fix alpha on rotate (for newer imagemagick versions)

chregu:
If the image doesn't have an alpha channel already, newer ImageMagick versions seem not to add that automatically anymore, even when the Pixel has an alpha channel. This fix adds an alphachannel to an image, if there was none before

### DIFF
--- a/lib/Imagine/Imagick/Image.php
+++ b/lib/Imagine/Imagick/Image.php
@@ -258,7 +258,11 @@ final class Image extends AbstractImage
 
         try {
             $pixel = $this->getColor($color);
-
+            // if the input image doesn't have an alpha channel, we need to add one
+            //  ImageMagick 7 (?) seems not to do that automatically anymore.
+            if ($this->imagick->getImageAlphaChannel() == \Imagick::ALPHACHANNEL_UNDEFINED) {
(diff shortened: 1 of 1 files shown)


lib/Imagine/Imagick/Image.php
```diff
@@ -258,7 +258,11 @@ public function rotate($angle, ColorInterface $background = null)
 
         try {
             $pixel = $this->getColor($color);
-
+            // if the input image doesn't have an alpha channel, we need to add one
+            //  ImageMagick 7 (?) seems not to do that automatically anymore.
+            if ($this->imagick->getImageAlphaChannel() == \Imagick::ALPHACHANNEL_UNDEFINED) {
```
romainneutron: Could you use strict comparison `===` ?

mlocati:
What about implementing what @romainneutron suggests and adding a test?

mlocati:
I think that the problem was in the `setColorspace` method of the Imagick driver.
This issue should be fixed in #673.
Please reopen this pull request (or another issue) if this problem still persists.

mlocati:
#673 was a wrong approach.
BTW I can't replicate this issue (see #674). If the problem persists please provide an example.

PR #559, opened 2017-05-09
move PHPUnit Bridge to dev requirements

craue:
it's only needed when running tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,11 +20,11 @@
         "bin-dir": "bin"
     },
     "require": {
-        "php": ">=5.3.2",
-        "symfony/phpunit-bridge": "^3.2"
+        "php": ">=5.3.2"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


aertmann:
Hey awesome came here to open such a PR! 😃

aertmann:
FYI needed this since `symfony/phpunit-bridge` is not compatible with PHPUnit 6 which is preferred with PHP 7 and they've only released a beta version that is compatible so far. Thus any installations requiring PHPUnit and Imagine running on PHP 7 break when running tests currently.

kdambekalns:
Awesome!

kornrunner:
Can this be released as a minor patch, please? Thanks!

romainneutron:
Version 0.7.1 has been released https://github.com/avalanche123/Imagine/releases/tag/v0.7.1

kornrunner:
Thank you!

PR #560, opened 2017-05-09
composer: require --dev symfony/phpunit-bridge ^3.2

kornrunner:
PHPUnit Bridge should be a development requirement. I think this change should be issued as a new version as current implementation can break things upstream.

Thanks!


robfrawley:
Duplicate of #559.

Issue #374, opened 2014-10-15
Unable to convert colorspace


wyred:
I have a jpg in CMYK colorspace.

Here's a simplified version of my code

```
$image = $imagine->open($source_file);
$image->crop($crop_point, $square_box)
    ->resize($resize_box)
    ->strip()
    ->usePalette(new Imagine\Image\Palette\RGB())
    ->save($destination_file);
```

But when I open the resulting file in photoshop, it's still in CMYK.

edigu:
@romainneutron is there any known issue exist which related to mac os x? I have same issue on os x.

trsteel88:
@edigu @romainneutron I'm having the same issue.

The dump says "rbg". However, saving the image does not result in an RBG image. After saving and downloading the file it is still CMYK.

trsteel88:
Using the command line it seems to work though.

![screen shot 2015-11-11 at 10 49 13 pm](https://cloud.githubusercontent.com/assets/869933/11091485/8d296f8a-88c6-11e5-89c8-78123cc31911.png)

So I am starting to this it is related directly to this library.

trsteel88:
Anyone?

ausi:
I think this bug gets fixed by #564

ausi:
I had the same issue, the problem was caused by a missing ImageMagick delegate called `lcms` or `lcms2`. You can see if this is installed on your system by calling `convert -list configure | grep DELEGATES` on the command line, if `lcms` doesn’t show up in this list you have to reinstall ImageMagick with the `--with-lcms2` or `--with-lcms` flag.

PR #564, opened 2017-05-16
Fix bug with colorspace conversion in Imagick

ausi:
On *macOS 10.11.6* using *Imagick 3.4.3, ImageMagick 6.9.5-2 Q16 x86_64 2016-07-13* I noticed that the `Imagine\Imagick\Image::usePalette()` method didn’t work as expected.

The unit test `Imagine\Test\Imagick\ImageTest::testUsePalette` fails on my machine with messages like *“Failed asserting that Imagine\Image\Palette\CMYK Object (...) is an instance of class "Imagine\Image\Palette\RGB".”*

After looking into it I found out that the issue can be fixed by using `Imagick::transformImageColorspace()`. After this change the tests are green. I also had to change the test code a little bit because it seems like ImageMagick has an optimization that converts gray RGB images to the grayscale colorspace automatically.

This is most likely the same issue as reported in #374 and I think it would get fixed by this pull request.


discordier:
It seems you have broken the `imagick` driver, as two tests are failing now or do the tests need adjustments?

ausi:
Thanks @discordier, I didn’t see the failing tests. I will look into it...

ausi:
By looking at the failing tests, I found the reason why `usePalette()` doesn’t work on my setup: The ImageMagick delegate lcms is not installed. This library is needed to get ICC color profiles working in ImageMagick.

The reason why it took me so long to find the cause for the problem is, that `Imagick::profileImage()` doesn’t throw an exception but fails silently.

I think there are three possible solutions for this in `Imagine\Imagick\Image::usePalette()`:

1. Switch to `Imagick::transformImageColorspace()` if lcms is not installed and log a warning that lcms is missing.
2. Switch to `Imagick::transformImageColorspace()` if lcms is not installed, but don’t trigger a warning message.
3. Throw an exception with a message similar to [Imagine/Gmagick/Image.php:699](https://github.com/avalanche123/Imagine/blob/1daff963d21474087f0a1fed2ffb830776eae31a/lib/Imagine/Gmagick/Image.php#L699).

This pull request currently implements option 1, I think this version is the best way to keep backwards compatibility on existing installs.

If desired, I can change this pull request to one of the other options.

robfrawley:
I would prefer throwing an exception if we cannot properly get the desired return value, like the `Gmagick` driver does.

ausi:
OK, then I will change the code to throw an exception. Should I keep the travis test runs that run without LCMS?

mlocati:
> The reason why it took me so long to find the cause for the problem is, that Imagick::profileImage() doesn’t throw an exception but fails silently.

I don't understand... You say that `profileImage` fails silently, but in order to check if it works you added this code:
```php
try {
    $image = new \Imagick();
    $image->newImage(1, 1, new \ImagickPixel('#fff'));
    $image->profileImage('icc', 'x');
}
catch (\ImagickException $exception) {
    // ...
}
```

So, will it fail silently or will it throw an exception?

ausi:
The check you mentioned works the other way round. It tries to use `Imagick::profileImage()` with an invalid profile. Now if `Imagick` throws an exception, we know that it actually tried to do something with the profile and thus generally supports color profiles. If it doesn’t throw an exception even though we passed an invalid profile, we know that it did not even look at the profile data and thus does not support color profiles.

ausi:
@mlocati should I update this pull request to reflect the changes requested in https://github.com/avalanche123/Imagine/pull/564#issuecomment-302276880 ?

mlocati:
Yes, I'd throw an exception. I'd also remove the test case, since it seems too much to me to have a new set of tests to be executed just to test if the environment is misconfigured...

lib/Imagine/Imagick/Image.php
```diff
@@ -858,6 +867,34 @@ private function detectColorspaceConversionSupport()
...
 
+    /**
+     * ImageMagick without the lcms delegate cannot handle profiles well.
+     * This detection is needed because there is no way to directly check for lcms.
+     *
+     * @return Boolean
+     */
+    private function detectProfilesSupport()
+    {
+        if (null !== static::$supportsProfiles) {
```
mlocati: I'd use `self` instead of `static` because everything is private. But it's just my personal taste: feel absolutely free to ignore this comment from me :wink:

ausi: Makes sense. Private members should be accessed via `self::`, done in ec96c8a463e18b81d0537414ae25604e0225595d

PR #565, opened 2017-05-16
Use the Imagick::IMGTYPE_...ALPHA constants if available

ausi:
The C constant `MagickCore::TrueColorMatteType` was renamed to `MagickCore::TrueColorAlphaType` in ImageMagick version 7.0.1, see [ImageMagick@def23e5](https://github.com/ImageMagick/ImageMagick/commit/def23e5d7331b1a13ed593b6d6aca516da382328). This change was added as `Imagick::IMGTYPE_TRUECOLORALPHA` in PHP Imagick 3.4.3, see [Imagick@87ed96d](https://github.com/mkoppanen/imagick/commit/87ed96d7998aa5fa6c6e14ac54167c410153da9e).

I think we should use the new `IMGTYPE_...ALPHA` constants because the `IMGTYPE_...MATTE` constants are deprecated.

### DIFF
--- a/lib/Imagine/Imagick/Image.php
+++ b/lib/Imagine/Imagick/Image.php
@@ -824,14 +824,12 @@ final class Image extends AbstractImage
         $typeMapping = array(
             // We use Matte variants to preserve alpha
             //
-            // (the constants \Imagick::IMGTYPE_TRUECOLORMATTE and \Imagick::IMGTYPE_GRAYSCALEMATTE do not exist anymore in Imagick 7,
-            // to fix this the former values are hard coded here, the documentation under http://php.net/manual/en/imagick.settype.php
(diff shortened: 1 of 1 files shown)


ausi:
It should already be used in 3.4.3, see https://github.com/mkoppanen/imagick/blob/3.4.3/imagick_helpers.c#L1182

But it seems the constant `Imagick::IMGTYPE_TRUECOLORALPHA` is only available if Imagick was compiled against ImageMagick 7.0.0 or higher, see https://github.com/mkoppanen/imagick/blob/3.4.3/imagick_helpers.c#L1179

mlocati:
:+1: in any case, supporting these new constants doesn't hurt :wink:

PR #573, opened 2017-10-18
Fix imagick rotation and canvas rezize, with tests

thePanz:
Updated PR with initial tests for rotation and canvas resize
See: https://github.com/avalanche123/Imagine/pull/526

### DIFF
--- a/lib/Imagine/Imagick/Image.php
+++ b/lib/Imagine/Imagick/Image.php
@@ -265,6 +265,8 @@ final class Image extends AbstractImage
             $pixel = $this->getColor($color);
 
             $this->imagick->rotateimage($pixel, $angle);
+            // Reset the image page: next operations will use the correct canvas size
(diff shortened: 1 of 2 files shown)


mlocati:
The test fails...

mlocati:
> Tests always fail when using imagik, but are successful when using gmagik (except for one case) :/

Well, yes: this PR is about imagick, not gmagick :wink:
I [fixed](https://github.com/avalanche123/Imagine/pull/573/commits/fb3ee1ede306c78d2fb177334d9dfd2ff58e0974) the gmagick failing case for PHP 5.3 (on PHP 5.3 you can't use `(new class())->method()`).

mlocati:
#thePanz Could you  provide an example when you have this issue? Please open another issue (or a pull request with a failing test case) with the required steps to reproduce it.

PR #576, opened 2017-11-10
Required for PHP 5.3

goncons:
PHP-images is NOT compatible with PHP 5.3, Precise is required.

See: https://docs.travis-ci.com/user/reference/trusty#PHP-images

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ script:
 
 php:
   - 5.3
+    dist: precise
   - 5.4
   - 5.5
   - 5.6


avalanche123:
I don't think this is valid yaml, in the examples from the linked article,  the `dist:` key is specified at the same level as `php:`

mlocati:
Since https://github.com/avalanche123/Imagine/pull/581 got merged, I think this PR is no more needed and can be closed

mlocati:
#581 resolved this issue

PR #587, opened 2018-02-26
Update composer.json

andreybolonin:


mlocati:
Imagine **does** support PHP versions as old as 5.3: why should it require PHP 5.6 in composer.json?

mlocati:
I'm closing this: if Imagine will require/use features on recent PHP versions we'll drop support to old PHP versions.

PR #588, opened 2018-02-26
Update .travis.yml

andreybolonin:


mlocati:
Imagine **does** support PHP versions as old as 5.3: why should it require PHP 5.6 in composer.json?

matthewtrask:
because 5.3 is dead and not getting security patches. The max this lib should support is 5.6, and that loses sec support this year.

mlocati:
> because 5.3 is dead and not getting security patches. The max this lib should support is 5.6, and that loses sec support this year.

And what should I do with one of our projects that runs on PHP 5.5, required us years of development, is mission critical for us and is not exposed on a public server?

matthewtrask:
... you upgrade it like everyone else in the world?

mlocati:
> ... you upgrade it like everyone else in the world?

Are you so sure?

matthewtrask:
More sure then you are apparently.

mlocati:
I'm really sure that there are a lot of companies that use unsupported PHP versions (the one I'm currently working for for instance, and no, they won't upgrade php).

In my 27 years of experience in IT, I have worked in many different environments (from one-man companies to companies with thousands of employees), with many different approaches, and I've learnt that  every system has his own rules.

robfrawley:
I also support removing EOF versions of PHP from this project. It provides two important benefits:
- As members of the PHP community, and the open source community in general, we have an obligation to motivate consumers of our projects to use secure interpreters. Doing so makes the web safer in general, for everyone. Furthermore, it helps promote progression of minimum PHP versions for all projects. No one should be using EOL PHP versions. It's not like the PHP project didn't provide extensive notice prior to ending support.
- Upgrading our minimum PHP version also allows us to take care of new language constructs and streamline the code.

> And what should I do with one of our projects that runs on PHP 5.5

In this case you stick with the last compatible release. Upping PHP requirements isn't a BC break. See: https://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html#why-dropping-php-support-in-a-minor-version-is-not-a-bc-break

> I'm really sure that there are a lot of companies that use unsupported PHP versions (the one I'm currently working for for instance, and no, they won't upgrade php).

In this case you would have to give them three options: 1. Upgrade PHP; 2. Write your own manipulation toolkit for images; or 3. Maintain their own fork of this bundle to handle any changes they want or to merge upstream changes back. Most sane companies would choose number 1.

mlocati:
Small companies with huge projects will choose 3.

matthewtrask:
Small companies don't have the bandwidth to fork and maintain a copy of a library. Those businesses hope sane maintainers are making the right decisions.

mlocati:
Exactly, but if package maintainers elect themselves as the holders of the Truth and the Guardians of the Perfect World where everybody uses the most recent PHP, point 3 is the only choice

matthewtrask:
Clearly you are holding on to some sort of fantasy of supporting old ass versions of PHP for no reason other then more work for you. Ill fork and bump up the spec to supported versions and keep it moving. Thanks for the motivation.

mlocati:
As I already wrote, the choice of keep using an unsupported PHP version is not mine (sadly). I really would like to switch to PHP 7.x for a number of reason...

robfrawley:
> [...] if package maintainers elect themselves as the holders of the Truth and the Guardians of the Perfect World where everybody uses the most recent PHP [...] -@mlocati

Let's not act like dropping `5.4` and `5.5` (upping the project minimum requirements to `5.6`) is anything close to a requirement that "everybody uses the most recent PHP". It is more akin to "requiring everybody uses active, supported version of PHP", the last of which (`5.6`) was release over four years ago and is effectively a drop-in replacement for the former releases; there are no notable or significant incompatible changes between `5.6` and `5.5` ([see documentation](http://php.net/manual/en/migration55.incompatible.php)) or `5.4` ([see documentation](http://php.net/manual/en/migration54.incompatible.php)).

Also, I would argue that this PR is rather conservative: some of the largest PHP projects require PHP `5.6` and many now require `7.x`:
- [Imagine Bundle (requires `7.1`)](https://github.com/liip/LiipImagineBundle/blob/2.0/composer.json#L21)
- [Symfony (requires `7.1`)](https://github.com/symfony/symfony/blob/master/composer.json#L19)
- [Laravel (requires `7.1`)](https://github.com/laravel/laravel/blob/master/composer.json#L8)
- [Monolog (requires `7.0`)](https://github.com/Seldaek/monolog/blob/master/composer.json#L16)
- [Twig (requires `7.0`)](https://github.com/twigphp/Twig/blob/2.x/composer.json#L30)
- [PHPUnit (requires `7.1`)](https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L24)
- [Doctrine (requires `7.1`)](https://github.com/doctrine/doctrine2/blob/master/composer.json#L19)
- [CakePHP (requires `5.6`)](https://github.com/cakephp/cakephp/blob/master/composer.json#L31)
- [OAuth-2 Server (requires `7.0`)](https://github.com/thephpleague/oauth2-server/blob/master/composer.json#L7)
- [Zend Framework (requires `5.6`)](https://github.com/zendframework/zendframework/blob/master/composer.json#L13)

Moreover, it's not like this change would make this project inacceccable to those running unsupported versions of PHP: you'd still be able to use prior versions and handle that however you'd like. 

With all this said though, this is only a discussion at this point and no dedision has been made one way or the other, so lets not use hyperbole: the end of the world is quite a ways off. :-)

> Clearly you are holding on to some sort of fantasy of supporting old *** versions of PHP for no reason other then more work for you. -@matthewtrask

Please keep discussion civil and "G" rated.

mlocati:
> Also, I would argue that this PR is rather conservative: some of the largest PHP projects require PHP 5.6 and many now require 7.x:

Those libraries require PHP 5.6+ because they use features of PHP 5.6+.
At the moment Imagine can run on PHP 5.3, and forcing a greater PHP version just "because you should use PHP 5.6+" doesn't seem correct to me: decision makers and developers are two distinct figures...

ausi:
Just as a side note: Some frameworks listed above still have active branches that support older PHP versions. For example Symfony 2.8 has its end of life in 11/2019 still supports PHP 5.3.

romainneutron:
Let's not drop support except if it brings direct benefit to the library itself. I don't see benefit in the listed arguments above.

andreybolonin:
https://wiki.php.net/todo/php73 php 7.3 is coming up...

mlocati:
I'm closing this: if Imagine will require/use features on recent PHP versions we'll drop support to old PHP versions.

PR #590, opened 2018-04-11
Use memory stream instead of output buffering

trowski:
This PR avoids using the output buffer to capture GD output by writing image data to an in-memory stream. Not terribly different, but cleaner IMO and avoids any possible issues surrounding output buffering.

### DIFF
--- a/lib/Imagine/Gd/Image.php
+++ b/lib/Imagine/Gd/Image.php
@@ -254,10 +254,13 @@ final class Image extends AbstractImage
      */
     public function get($format, array $options = array())
     {
-        ob_start();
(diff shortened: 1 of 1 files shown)


mlocati:
I tested this with imagegif:

- with PHP 5.3 we have `imagegif() expects parameter 2 to be string, resource given`
- with PHP 5.4 and 5.5 we `rewind(): 6 is not a valid stream resource` (I think because `imagegif` closes the stream)
- with PHP 5.6+ it works

I'm going to close this, BTW if/when we'll drop support for old PHP versions we can apply this patch.

PR #504, opened 2016-08-03
Support webp 

chregu:
These lines are needed to support webp with the imagick adapter.

I don't intend to write webp support for the other adapters, but this may be a start for someone else.

Also requested in #271

PR #591, opened 2018-04-24
Remove not-supported exception for image option

ausi:
There are some image options that are only supported by some adapters and in the past these always got ignored.

With #504 a `NotSupportedException` was added for the `webp_lossless` option in the Gmagick adapter (but not the GD adapter). That was probably a mistake? If so, I think we should remove the exception.

/cc @chregu @alexander-schranz

### DIFF
--- a/lib/Imagine/Gmagick/Image.php
+++ b/lib/Imagine/Gmagick/Image.php
@@ -287,10 +287,6 @@ final class Image extends AbstractImage
             $image->setCompressionQuality($options['webp_quality']);
         }
 
-        if (isset($options['webp_lossless']) && in_array($format, array('webp'))) {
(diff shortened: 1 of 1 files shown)


alexander-schranz:
As mentioned in https://github.com/liip-forks/Imagine/pull/3 I did oriented at the coalesce 
 function which does also throw a NotSupportedException: https://github.com/avalanche123/Imagine/blob/68d416ee2be3423c5758bddd31be11fcd80d16e5/lib/Imagine/Gmagick/Layers.php#L76. But for me its ok when we remove it.

ausi:
As far as I understand the codebase, unsupported methods should throw, but unsupported options should not (e.g. `resolution-units`, `resampling-filter`, `flatten`,…).

robfrawley:
@ausi: Your logic is reasoned.

PR #598, opened 2018-07-18
Catch exif_read_data exceptions

lentex:
Instead of throwing exceptions
when exif_read_data fails, just
catch the exception and return
empty array.


jaylinski:
@avalanche123 can we get this merged as soon as possible? 🙏

PR #605, opened 2018-08-16
Fix URL of remote test images

mlocati:


dmarkowicz:
GraphicsMagick supports only http:// ftp:// file:// urls via libxml2, which does not support HTTPS.
see: GraphicsMagick-1.3.30/coders/url.c

mlocati:
#607 is a much better approach

Issue #415, opened 2015-03-04
fopen vs curl for external images


glend:
Most shared hosting providers disable the setting allow_furl_open, therefore you can't fetch external images. There are ways to override this setting in certain providers but they highly discourage it because of security issues?

Why not use curl instead?

romainneutron:
I'd gladly appreciate we refactor this part of the app to allow such feature, but it's not on the roadmap

mlocati:
Implemented in #607

PR #607, opened 2018-08-17
Add File\LoaderInterface with basic curl support

mlocati:
What about adding a simple loader that can fetch remote files with curl?
This would solve #415, and fix the problem that GraphicsMagick can't load images from https.


lib/Imagine/File/Loader.php
```diff
@@ -0,0 +1,295 @@
...
+    {
+        if (is_object($path) && !method_exists($path, '__toString')) {
+            throw new InvalidArgumentException(sprintf('$path is an object of file %s which does not implement the __toString method', get_class($path)));
+        }
+
+        $this->path = (string) $path;
+        if ($this->path === '') {
+            throw new InvalidArgumentException('$path is empty');
+        }
+        $this->isUrl = filter_var($this->path, FILTER_VALIDATE_URL);
```
dmarkowicz: Should be casted to bool, now url is assigned.

tests/Imagine/Test/Image/AbstractImagineTest.php
```diff
@@ -116,7 +116,7 @@ public function testShouldCreateImageFromString()
 
     public function testShouldCreateImageFromStreamWithMetadata()
     {
-        $source = 'http://imagine.readthedocs.org/en/latest/_static/exit-90-test.jpg';
+        $source = 'https://imagine.readthedocs.org/en/latest/_static/exit-90-test.jpg';
```
dmarkowicz: Most likely you want to test it in testShouldOpenAnHttpImage() and change HTTP_IMAGE to https. Here image is opened by `fopen`.

mlocati: I'll revert this edit. About changing the HTTP_IMAGE constant: it should work even if we leave it with the HTTP protocol (the standard CURL adapter is configured to follow redirects).

mlocati:
I'm going to merge this because it's required to let tests pass (it's blocking other pull requests).

PR #610, opened 2018-08-17
Add coding style rule checks with php-cs-fixer

### LINKED ISSUE #606: PHP-CS-Fixer
What about normalizing the coding style by using [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)?

I'm thinking about these rules (you can (re)configure them by importing them into [php-cs-fixer-configurator](https://mlocati.github.io/php-cs-fixer-configurator/?version=2.2#configurator)):

```yaml
version: 2.2.20
expandSets: true
fixerSets:
  - '@Symfony'
fixers:
  array_syntax: true
  blank_line_before_return: false
  dir_constant: true
  ereg_to_preg: true
  function_to_constant: true
  is_null:
    use_yoda_style: false
  no_alias_functions: true
  no_php4_constructor: true
  no_short_echo_tag: true
  no_unreachable_default_argument_value: true
  no_useless_return: true
  non_printable_character: true
  ordered_imports: true
  pre_increment: false
  phpdoc_add_missing_param_annotation:
    only_untyped: false
  phpdoc_align: false
  phpdoc_order: true
  protected_to_private: false
  psr4: true
risky: true
```

mlocati:
Closes #606

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ docs/_build/cache
 composer.lock
 vendor/
 bin/
+/.php_cs.cache
+/.php_cs
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,20 +28,21 @@ install:
             PHPUNIT_EXCLUDED_GROUPS=none
             if test "${IMAGINE_DRIVER}" = 'imagick'; then ./.travis/imagick.sh; else PHPUNIT_EXCLUDED_GROUPS="${PHPUNIT_EXCLUDED_GROUPS},ext-imagick"; fi
             if test "${IMAGINE_DRIVER}" = 'gmagick'; then ./.travis/gmagick.sh; else PHPUNIT_EXCLUDED_GROUPS="${PHPUNIT_EXCLUDED_GROUPS},ext-gmagick"; fi
+            composer --no-interaction remove --dev --no-update --no-scripts friendsofphp/php-cs-fixer
             php --ri curl || true
             ;;
     esac
-  - composer install
+  - composer --no-interaction install
 
 script:
   - |
     set -o errexit
     case "${TRAVISOPERATION:-}" in
         codestyle)
-            composer codestyle
+            composer --no-interaction codestyle
             ;;
         *)
-            composer test -- --exclude-group "${PHPUNIT_EXCLUDED_GROUPS}"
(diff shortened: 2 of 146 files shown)


.travis.yml
```diff
@@ -45,3 +32,39 @@ matrix:
...
+
+before_install:
+  - phpenv config-rm xdebug.ini || true
+
+install:
+  - |
+    set -o errexit
+    case "${TRAVISOPERATION-}" in
+        codestyle)
+            cp .travis/composer.json.codestyle ./composer.json
```
alexander-schranz: @mlocati I would recommend of running `composer require friendsofphp/php-cs-fixer:^2.12 --dev` instead of copy composer.json or using [styleci](https://styleci.io) instead.

.gitignore
```diff
@@ -4,3 +4,4 @@ docs/_build
 composer.lock
 vendor/
 bin/
+.php_cs.cache
```
alexander-schranz: missing also `/.php_cs`

.travis.yml
mlocati: I love StyleCI, but it not as configurable as php-cs-fixer. Furthermore AFAIK StyleCI doesn't support reading from `.php_cs`, so we'd need to:
- maintain the StyleCI configuration
- maintain a php-cs-fixer configuration so that developers can check/format the PHP code locally before submitting pull requests.

About using `composer require friendsofphp/php-cs-fixer:^2.12 --dev` instead of a separate `composer.json` file will increase the test time, since composer will have to install all the default packages, even if we just need `php-cs-fixer` since it's the only tool required.

.gitignore
mlocati: PR updated

.travis.yml
alexander-schranz: @mlocati then I would more go for the global installation over: `wget https://cs.sensiolabs.org/download/php-cs-fixer-v2.phar -O php-cs-fixer` of the php-cs-fixer.
and as long as php-cs-fixer is not in the require-dev of the normal composer.json the contributers will not be familiar to use it.

mlocati: I agree. I updated this pull request so that developers can run php-cs-fixer simply with `composer codestyle`

PR #612, opened 2018-08-17
[wip] Unify magics

mlocati:
This is just a test to see if we can test imagick and gmagick together in TravisCI, so that we can reduce the total test time.

Since #607 is required to fix tests, this PR is based on it (that's why it's a work in progress).


mlocati:
Dropping this PR since tests hung on PHP 5.3

PR #630, opened 2018-08-22
Implement customizable factory methods

mlocati:
What about introducing a FactoryClassInterface class that developers can implement in order to override the default classes created?

Currently, the classes that can be implemented are:
- Imagine\Image\Metadata\MetadataReaderInterface
- Imagine\Image\BoxInterface
- Imagine\Image\FontInterface
- Imagine\File\LoaderInterface
- Imagine\Image\LayersInterface
- Imagine\Image\ImageInterface

Supersedes #212 and #279

PR #632, opened 2018-08-22
Implement customizable factory methods - V2

mlocati:
An alternative to #630

### DIFF
--- a/lib/Imagine/Factory/ClassFactory.php
+++ b/lib/Imagine/Factory/ClassFactory.php
@@ -2,19 +2,19 @@
 
 namespace Imagine\Factory;
 
-use Imagine\Image\Metadata\ExifMetadataReader;
-use Imagine\Image\Metadata\DefaultMetadataReader;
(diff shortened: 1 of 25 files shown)


lib/Imagine/Factory/ClassFactory.php
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * {@inheritdoc}
+     *
+     * @see \Imagine\Factory\ClassFactoryInterface::createDrawer()
+     */
+    public function createDrawer($handle, $resource)
+    {
+        switch ($handle) {
+            case static::HANDLE_GD:
```
stof: you should use `self::` instead of relying on late static binding. Overriding the constants is not a supported extension point anyway (as the code is likely to use the ones from the interface)

mlocati: 👍 done in f00fe304343f0657ce51a80275c7bb227afa57a7

PR #637, opened 2018-08-25
Move code from /lib/Imagine to /src and update APIs

### LINKED ISSUE #635: Moving code from lib/Imagine to src?
Imagine code is stored in the `lib/Imagine` folder, whereas libraries usually use `/src`.
So, what about switching to that de facto standard?

mlocati:
Closes #635


Koc:
This will break all other PRs. Is it better review other PRs first?

mlocati:
I reviewed all the other pull requests already, and I'm awaiting feedback.
In any case, git should be smart enough to follow the renames (indeed in the pull requests you can read "This branch has no conflicts with the base branch.")

PR #642, opened 2018-08-27
Fix annotations of tests

mlocati:


mlocati:
This is just a cosmetic change, to avoid phpunit considering tests as *risky*.

For example, with IMAGINE_DRIVER=imagick and php 7.2 we have:
- previous phpunit results: `Tests: 523, Assertions: 978, Skipped: 29, Risky: 40.`
- phpunit results with this PR: `Tests: 545, Assertions: 987, Skipped: 28.`

PR #645, opened 2018-08-27
Add THUMBNAIL_UPSCALE setting - Rebased

mlocati:
This is #345 rebased


tests/Imagine/Test/Image/AbstractImageTest.php
```diff
@@ -284,14 +284,22 @@ public function testThumbnailShouldReturnACopy()
         $this->assertNotSame($image, $thumbnail);
     }
 
-    public function testThumbnailWithInvalidModeShouldThrowAnException()
+    public function testThumbnailWithInvalidSettingShouldThrowAnException()
     {
         $factory = $this->getImagine();
         $image = $factory->open('tests/Imagine/Fixtures/google.png');
         $this->{method_exists($this, $_ = 'expectException') ? $_ : 'setExpectedException'}('Imagine\Exception\InvalidArgumentException', 'Invalid mode specified');
```
vlakoff: You missed a "s/mode/setting" here :)

mlocati: @vlakoff Thanks. BTW any reason to rename the `$mode` argument to `$setting`? I think that this argument still defines the resize *mode*...

vlakoff: That's because it is broader.

Previously, we could only choose between one mode or the other.

Now we can specify one or several settings, combining them. One mode or the other, and an option to allow upscaling (which isn't a mode).

lib/Imagine/Image/AbstractImage.php
```diff
@@ -63,33 +68,60 @@ public function thumbnail(BoxInterface $size, $mode = ImageInterface::THUMBNAIL_
...
+
+        $allSettings = ImageInterface::THUMBNAIL_INSET | ImageInterface::THUMBNAIL_OUTBOUND | ImageInterface::THUMBNAIL_UPSCALE;
+
+        if (!is_int($settings) || ($settings & ~$allSettings)) {
+            throw new InvalidArgumentException('Invalid setting specified');
+        }
+
+        if ($settings & ImageInterface::THUMBNAIL_INSET &&
+            $settings & ImageInterface::THUMBNAIL_OUTBOUND) {
+            throw new InvalidArgumentException('Only one mode should be specified');
```
vlakoff: Michele, see also here. By using "settings", we can distinguish "all settings" / "only modes".

tests/Imagine/Test/Image/AbstractImageTest.php
mlocati: BTW it's main purpose still hasn't changed: it still defines how the image should be resized, isn't it?
If so, IMHO a parameter that describes how to do something is still a mode...

lib/Imagine/Image/AbstractImage.php
mlocati: I simply think that the word *mode* gives much more relevance to the argument than *setting*. Setting seems like an optional stuff, mode highlights its relevance. BTW I'm not going to waste time on this subject :wink:

vlakoff:
I understand your point of view, but I still prefer mine. For me, "settings" are not forcibly optional and can be something required.

Let's say both names would be fine. Otherwise, by 2021 (approx.) this still won't be merged ^^

lib/Imagine/Image/AbstractImage.php
```diff
@@ -63,33 +68,60 @@ public function thumbnail(BoxInterface $size, $mode = ImageInterface::THUMBNAIL_
...
+        }
+
+        return $settings;
```
mlocati: @vlakoff Shouldn't $settings contain one of THUMBNAIL_INSET or THUMBNAIL_OUTBOUND?

vlakoff: In `thumbnails()`, the `$settings` parameter has a default value, and more relevant, it's actually supported to omit the mode, see my next comment.

vlakoff: I'm already handling all the probable errors, and the code is already quite complex. So I tried to not complexify the code more than needed.

This being said, `$settings` being a bitmask, a value of `0` is not that wrong, it just means "specify no flag". Still, I'd tend to agree this value is not supposed to be used. But I think adding code to reject it is not worth it.

That's because, how I thought the API, this 2nd parameter `$settings` is supposed to be "omittable" (calling the method with only the 1st parameter), but not "skippable" (if you use the 3rd parameter, you have to provide a meaningful value for this 2nd parameter). So I consider values like `0` or `null` to be unexpected and unsupported.

vlakoff: A solution could be to add a `THUMBNAIL_DEFAULT = 0` constant, that could be used to skip the parameter. But I think the added complexity is not worth it. I think it's better to require the user to specify a meaningful value for the 2nd parameter if he uses the 3rd parameter, and consider `0` as an unsupported value (i.e. works, but unexpected).

This being said, you are free to tweak the PR if you wish, I'm fine with it.

lib/Imagine/Image/ManipulatorInterface.php
```diff
@@ -22,8 +22,26 @@
...
 interface ManipulatorInterface
 {
-    const THUMBNAIL_INSET = 'inset';
-    const THUMBNAIL_OUTBOUND = 'outbound';
+    /**
+     * The original image is scaled so it is fully contained within the thumbnail dimensions (the image width/height ratio doesn't change).
+     *
+     * @var int
+     */
+    const THUMBNAIL_INSET = 0x00000001;
```
alexander-schranz: @mlocati This BC Break is undocumented and should be added to the CHANGELOG.md

mlocati: @alexander-schranz What are you referring to with "This BC Break"?

alexander-schranz: changing a constant value should be documented if your image configuration is not in php you can't use the constant and use the inset as string.

mlocati: You still can use the `'inset'` string: it's automatically converted to the integer value (see https://github.com/avalanche123/Imagine/pull/645/files#diff-6c8ebea98c0d67b0ee6c2083e4ab610aR114)

alexander-schranz: Yeah, but the change should still be documented as something like: `$mode !== ManipulatorInterface::THUMBNAIL_INSET` will not longer work.

mlocati: You are right! What about editing the CHANGELOG.md file yourself and submitting a pull request?

alexander-schranz: @mlocati I can in which version was this first released?

mlocati: This PR has been merged on 2018-08-28, and the first version after that date has been v1.0.0-alpha1
PS: reading the CHANGELOG.md, it seems that this change was already documented: see these lines:
> Add flag to thumbnail to allow upscaling images (@vlakoff)
> NOTE the $mode argument has been renamed to $settings, and it's now an integer (but old string values are accepted for backward compatibility)

PR #650, opened 2018-08-28
Remove final from classes and methods

### LINKED ISSUE #162: final classes

mlocati:
Closes #162


mlocati:
Imagine still requires changes in the API, so I'd prefer to postpone this to another version.

PR #651, opened 2018-08-29
Let ExifMetadataReader::doReadData always return a MetadataBag

mlocati:
ExifMetadataReader::extract can return an array or a MetadataBag instance.
In the phpdoc, we have that ExifMetadataReader::doReadData should return a MetadataBag instance: let's make it so.


mlocati:
This PR is wrong

mlocati:
The right fix is #654

PR #672, opened 2018-09-03
Fix Imagick save compression of PNG images

mlocati:
First, let's add a test case


mlocati:
Then, let's fix the issue

PR #675, opened 2018-09-04
Add getChannelsMaxValue to PaletteInterface

mlocati:
It's often useful to know the max value that a color channel can accept


mlocati:
I'm merging this right now because I need this for another pull request.

PR #691, opened 2018-10-23
Call setImageType/setImageColorspace when setting Imagick color space

mlocati:


mlocati:
This is wrong (it mangles colorspaces).

PR #700, opened 2018-12-05
Allows for extra metadata sections to be read

### LINKED ISSUE #568: [ExifMetadataReader] Omitted metadata sources such as GPS
Currently, [`ExifMetadataReader` reads from `EXIF`, `IFD0`](https://github.com/avalanche123/Imagine/blob/develop/lib/Imagine/Image/Metadata/ExifMetadataReader.php#L102).

But, [exif_read_data](http://php.net/manual/en/function.exif-read-data.php) provides: `FILE`, `COMPUTED`, `IFD0`, `THUMBNAIL`, `EXIF`, `GPS`, `INTEROP` and probably others. This means those other metadata sources end up being ignored (which is a problem for me as I was trying to figure out why GPS wasn't getting read properly).

You should probably allow for additional metadata sources here or at least allow for specifying sources in the constructor.

hermzz:
By setting them in the `ExifMetadataReader` constructor.

Fixes #568.

### DIFF
--- a/src/Image/Metadata/ExifMetadataReader.php
+++ b/src/Image/Metadata/ExifMetadataReader.php
@@ -115,14 +115,10 @@ class ExifMetadataReader extends AbstractMetadataReader
         }
 
         $metadata = array();
-        $sources = array('EXIF' => 'exif', 'IFD0' => 'ifd0');
 
-        foreach ($sources as $name => $prefix) {
-            if (!isset($exifData[$name])) {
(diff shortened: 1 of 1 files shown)


mlocati:
What about simply reading all the available EXIF data?

hermzz:
Yeah, why not? Makes more sense.

mlocati:
Superseded by #701

PR #706, opened 2019-03-13
Fix passing big integers to box constructor

### LINKED ISSUE #705: Box constructor fails on very high integer arguments
### Issue description

When passing integer values near PHP_INT_MAX as $width or $height argument to the `Box` constructor,
an exception is thrown: "Length of either side cannot be 0 or negative, current size is 500x9223372036854775807"

9223372036854775807 is PHP_INT_MAX on 64bit systems. It is clearly a positive integer and not negative. It turns out that the expression `(int) round($height)` used in the constructor method returns the negated value -9223372036854775808 ! So the `round()` function should only be used if the provided width and height values are not integers.

### What version of Imagine are you using?

1.2.1-dev

### What's the PHP version you are using?

7.2

### What's the imaging library you are using [gd/imagick/gmagick/any]?

any

### Minimal PHP code to reproduce the error: 

```php

$size = new \Imagine\Image\Box(PHP_INT_MAX, PHP_INT_MAX);

```

wasinger:
Fixes #705

### DIFF
--- a/src/Image/Box.php
+++ b/src/Image/Box.php
@@ -38,8 +38,14 @@ final class Box implements BoxInterface
      */
     public function __construct($width, $height)
     {
-        $this->width = (int) round($width);
-        $this->height = (int) round($height);
+        if (!\is_int($width)) {
+            $width = (int) round($width);
+        }
+        if (!\is_int($height)) {
+            $height = (int) round($height);
+        }
+        $this->width = $width;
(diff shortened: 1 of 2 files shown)


mlocati:
Could you fix https://travis-ci.org/avalanche123/Imagine/jobs/505837746#L629 ?

mlocati:
Thanks!

Issue #460, opened 2015-09-11
Masks not working as intended


aewing:
I have an image I've created using imagine->create()...
I am trying to copy another photo (via URL) into this image after applying a circular mask to it.
````php
      $size = new Box(160, 160);
      $photo = $this->imagine->load($raw_photo)->thumbnail($size, Image::THUMBNAIL_OUTBOUND);

      $mask = $this->imagine->create($size);
      $mask->draw()->ellipse(new Point\Center($size), $size, $this->palette->color('#000000'), 0);
      $photo->applyMask($mask);

      $img->paste($photo, new Point(10, 10));
````

The image displays normally, but the mask is ignored.

**Comment by aewing** [comment]
This was using the GD Imagine provider. Switching to Imagick has resolved the issue.

**Comment by tobias-r** [comment]
Hi, I have the same issue and I have a solution for it:

Replace code in Imagine\GD\Image.php in line 380 with:
```
                $delta     = (int) round($color->getAlpha() * $maskColor->getRed() / 255) * -1;

                if (false === imagesetpixel($this->resource, $x, $y, $this->getColor($color->dissolve($delta)))) {
                    throw new RuntimeException('Apply mask operation failed');
                }
```

The current code will always return "100" if the original image as an alpha of 100 at the point.

Would be great if you guys could integrate my code into production. It's tested!

ahukkanen:
I can confirm that with `gd` the image masks are still not working correctly. The solution suggestion provided by @tobias-r seems to fix the problem.

robfrawley:
@tobias-r @ahukkanen Would either of you care to implement a PR that addressed your concerns? That would be great!

mlocati:
IMHO calling `$photo->applyMask($mask->mask());` should fix the issue.

systemist:
Created a pull request containing @tobias-r suggested fix.

PR #708, opened 2019-04-16
applyMask() fix in Gd/Image

ninze:
Created applyMask() fix for #460


ausi:
I’m not sure if this fix is correct. I think `applyMask()` is meant to be used with a grayscale image.

Did you try to use `->applyMask($mask->mask());` as described in https://github.com/avalanche123/Imagine/issues/460#issuecomment-425042314 ?

systemist:
@ausi yes I did without success. The fix makes imagine work consistently with both GD and Imagick drivers. We are using this fix in production since April without any issues.

mlocati:
Is it possible to have a test case that covers what this PR is fixing?

ninze:
It probably is, will look into it.

ninze:
Done. I don't have gmagick installed, but GD and Imagick produce identical results after the fix.

jakubjo:
> @ausi yes I did without success. The fix makes imagine work consistently with both GD and Imagick drivers. We are using this fix in production since April without any issues.

Maybe it's about time someone merges this?

ninze:
After more than 2 years since the pull request was created, I also think someone should either merge or close this.

ausi (approved):
Tested this pull request locally and all worked correctly.

In the old code `->dissolve(…)` could have never ended up with a negative value and therefore could never have masked anything.

mlocati:
Thanks for reviewing this PR, @ausi !

chekalsky:
@mlocati thank you! is it possible to release new version of Imagine?

mlocati:
@chekalsky I'd publish a new version after tests have been fixed (I'd move them to GotHub Actions)

Issue #636, opened 2018-08-25
Replace Adobe license PDF files with plain text files?


mlocati:
This library contains Adobe license files in PDF format.
This is a waste of space: what about replacing them with plain text files?

mlocati:
Fixed by https://github.com/avalanche123/Imagine/commit/19a5897b672940b6b2e77fd6ddf04f6535abc912

PR #710, opened 2019-04-17
convert pdf files to plain text. See #636

garak:


mlocati:
Thanks! BTW I merged https://github.com/avalanche123/Imagine/commit/19a5897b672940b6b2e77fd6ddf04f6535abc912

I wrote that commit some time ago, but I didn't have time to read all the licenses until now (it seems there's no restrictions about the format of the license files).

Issue #501, opened 2016-06-21
Support for pdf, tif, jp2, webp


asajedi:
Hi,
is it possible to add support for the following formats:
- pdf
-tif
- jp2
- webp

in 
lib/Imagine/Imagick/Image.php

    private function getMimeType($format)
    {
        static $mimeTypes = array(
            'jpeg' => 'image/jpeg',
            'jpg'  => 'image/jpeg',
            'gif'  => 'image/gif',
            'png'  => 'image/png',
            'wbmp' => 'image/vnd.wap.wbmp',
            'xbm'  => 'image/xbm',
        );

        if (!isset($mimeTypes[$format])) {
            throw new RuntimeException(sprintf('Unsupported format given. Only %s are supported, %s given', implode(", ", array_keys($mimeTypes)), $format));
        }

        return $mimeTypes[$format];
    }

antoligy:
Might be worth adding this into #504 ?

theDyingMountain:
The Imagine Classes still dont't support webp. E.g. imagecreatefromstring is called, which doesn't support webp see [https://github.com/avalanche123/Imagine/blob/develop/src/Gd/Imagine.php#L91](https://github.com/avalanche123/Imagine/blob/develop/src/Gd/Imagine.php#L91)

lashus:
I'm surprised that for 3 years this hasn't been resolved yet :) I've created PR for this - @avalanche123  could you please take a look? It's not a BC.

mlocati:
> I'm surprised that for 3 years this hasn't been resolved yet :)

Maybe you are the first one that needs it and is willing to implement it :wink:

lashus:
> Maybe you are the first one that needs it and is willing to implement it

Don't take me wrong - I wasn't trying to be negative :) Just an observation and clearly the webp format is not as popular as google thinks ^^

PR #711, opened 2019-05-09
Webp full support

lashus:
According to #501


src/Gd/Imagine.php
```diff
@@ -239,4 +249,19 @@ private function withoutExceptionHandlers($callback)
...
     }
+
+    private function isWebp(string $data)
+    {
+        return strncmp( substr( $data, 8, 7 ), "WEBPVP8", 7 ) === 0;
+    }
+
+    private function loadWebp(string $data)
+    {
+        $tmpfile = tempnam(sys_get_temp_dir(), 'imaginewebp_');
```
mlocati: I think that:
1. The temporary directory should be configurable
2. We should delete the temporary file once we've done reading it

lashus: I'm not sure if that's necessery. Why would we want to configure the temp directory for just loading the file? 

Also if we want to delete the file we should probably do this somewhere else - we give the resource back and we cannot remove file that is currently opened as resources AFAIK. Perhaps on destructor of the class? To be honest - if it's in /tmp dir then filesystem will clear it eventually so it seemed like a "safe" choice :)

mlocati: I've seen crappy hosting providers where the directory returned from `sys_get_temp_dir()` was not writable...

About deleting the file: PHP loads the file content in memory, so the file can be safely deleted.

src/Gd/Imagine.php
```diff
@@ -87,9 +87,14 @@ public function open($path)
         $path = $loader->getPath();
 
         $data = $loader->getData();
-        $resource = $this->withoutExceptionHandlers(function () use (&$data) {
-            return @imagecreatefromstring($data);
-        });
+
+        if($this->isWebp($data)) {
```
mlocati: `imagecreatefromwebp` has been introduced in PHP 5.4, but Imagine [supports PHP 5.3.2](https://github.com/avalanche123/Imagine/blob/1.2.0/composer.json#L20). so what about

```suggestion
        if(function_exists('imagecreatefromwebp') && $this->isWebp($data)) {
```

src/Gd/Imagine.php
```diff
@@ -208,9 +213,14 @@ private function requireGdVersion($version)
      */
     private function doLoad($string, MetadataBag $metadata)
     {
-        $resource = $this->withoutExceptionHandlers(function () use (&$string) {
-            return @imagecreatefromstring($string);
-        });
+
+        if($this->isWebp($string)) {
```
mlocati: `imagecreatefromwebp` has been introduced in PHP 5.4, but Imagine [supports PHP 5.3.2](https://github.com/avalanche123/Imagine/blob/1.2.0/composer.json#L20). so what about

```suggestion
        if(function_exists('imagecreatefromwebp') && $this->isWebp($data)) {
```

adriancodesailors: Well, you have to choose whether you want to support version that is 10 years old and unsupported since last 5 years or to use the image standard from 2010, early adopted by facebook & google in 2013 :) I'm all up for the latter but I can work on my fork so it's up to you.

src/Gd/Imagine.php
adriancodesailors: Ok, will fix that in next commit after we resolve the second issue.

lashus: Ok, will fix that in next commit after we resolve the second issue.

src/Gd/Imagine.php
adriancodesailors: Well, you have to choose whether you want to support version that is 10 years old and unsupported since last 5 years or to use the image standard from 2010, early adopted by facebook & google in 2013 :) I'm all up for the latter but I can work on my fork so it's up to you.

lashus: Well, you have to choose whether you want to support version that is 10 years old and unsupported since last 5 years or to use the image standard from 2010, early adopted by facebook & google in 2013 :) I'm all up for the latter but I can work on my fork so it's up to you.

Do you think that function check is good enough? Shouldn't we post some information that with the newer version of PHP this will be working at least?

(Sorry for the other comment, wrong account).

mlocati: I'd just do the function exists check. We can add a note to the changelog / release notes that webp support requires PHP 5.4+.

lashus: I've pushed the new commit with your suggestion in it :)

mlocati:
Tests are failing.
Also note the [coding style issues](https://travis-ci.org/avalanche123/Imagine/jobs/530723134).

lashus:
> Tests are failing.
> Also note the [coding style issues](https://travis-ci.org/avalanche123/Imagine/jobs/530723134).

I've fixed code styling and tests. Do you want me to rebase as well? The Appveyor tests are failing but not because of my code :( There are a lot of issues with imagick from what I can tell.

ausi:
I created an alternative pull request for this issue: #718

Instead of using a temp directory, it uses a data URI to load the WEBP image.

Maybe this would be a better solution? We often had issues with using `sys_get_temp_dir()` in the past.

src/Gd/Imagine.php
```diff
@@ -239,4 +264,31 @@ private function withoutExceptionHandlers($callback)
...
     }
+
+    /**
+     * @param string $data
+     *
+     * @return bool
+     */
+    private function isWebp($data)
+    {
+        return 0 === strncmp(substr($data, 8, 7), 'WEBPVP8', 7);
```
mlocati: Isn't this the same as `substr($data, 8, 7) === 'WEBPVP8'`?

Issue #709, opened 2019-04-16
rotate() inconsistencies between drivers


systemist:
### Issue description

Gd and Imagick drivers leave the image in a different state after rotation. 

### What version of Imagine are you using?

1.2.1-dev

### What's the PHP version you are using?

PHP 7.2.1

### What's the imaging library you are using [gd/imagick/gmagick/any]?

Imagick, Gd

### Minimal PHP code to reproduce the error: 

```php

$size = $image->getSize();
$image->rotate(45);
$image->crop(new Point(0, 0), $size);

```

Output by Gd driver:
![Gd-output](https://user-images.githubusercontent.com/7146670/56239037-e330fa00-6087-11e9-9d5f-b6afc8510b0a.jpeg)

Output by Imagick driver:
![Imagick-output](https://user-images.githubusercontent.com/7146670/56239061-f0e67f80-6087-11e9-9b5c-6bbca80c4271.jpeg)

The problem is that Imagick does not update image geometry after the rotation. I think a similar issue is described here: https://phpimagick.com/Tutorial/imageGeometryReset

Output from Gd seems to be the correct one.

The solution would be to call `setImagePage()` inside Imagick driver after the rotation.

mlocati:
Fixed by #734

PR #712, opened 2019-05-18
Issue 709

lashus:
I've changed the cropping behaviour as written in #709. Not sure though how to test it and also without using distort and SRT we are changing the image size (in GD that doesnt happen). Not sure how to refactor it other way though as with using image distortion with SRT it's hard to rotate it in same way (using same point).


src/Imagick/Image.php
```diff
@@ -304,7 +304,11 @@ public function rotate($angle, ColorInterface $background = null)
         try {
             $pixel = $this->getColor($background);
 
-            $this->imagick->rotateimage($pixel, $angle);
+            $imageWidth = $this->imagick->getImageWidth();
+            $imageHeight = $this->imagick->getImageHeight();
+
+            $this->imagick->rotateImage($pixel, $angle);
+            $this->imagick->setImagePage($imageWidth, $imageHeight, 0, 0);
```
ausi: Setting the image page to the dimensions before the resize seems to be wrong because the image size grows on rotation.

From the [ImageMagick documentation](https://imagemagick.org/script/command-line-options.php#repage): 
> A given a canvas size of zero such as '0x0' forces it to recalculate the canvas size so the image

So `->setImagePage(0, 0, 0, 0);` should do the job here.

ausi:
> Not sure though how to test it

You can add a test like the following to `AbstractImageTest` class:

```php
public function testRotateWithCrop()
{
    $palette = new RGB();
    $color = $this
        ->getImagine()
        ->create(new Box(100, 100), $palette->color('#f00'))
        ->rotate(45, $palette->color('#fff'))
        ->crop(new Point(0, 0), new Box(100, 100))
        ->getColorAt(new Point(0, 50));
    $this->assertSame('#ffffff', (string) $color);
}
```

It creates as red 100x100 square image rotates it by 45 degrees and crops it to 100x100. After that the center pixel on the left side should be white as you can see in the following example images:

wrong: wrong correct: correct

> we are changing the image size (in GD that doesnt happen)

I don’t understand what you mean by that. I tested all three drivers and (with the fix for Imagick) they all behave the same. GD also increases the image size if you rotate the image by 45 degrees.

***

Your pull request seems to include unrelated commits. I think you meant to include just the commit 58193ee8c7b6aa8212f75f7212f530e5f4732797 correct?

ausi:
I created a followup pull request that includes the changes from my comments: #734

mlocati:
Superseded by #734
Please don't try to fix multiple stuff in the same pull request

PR #713, opened 2019-05-22
Fix undefined method error with older PHP versions

ahukkanen:
The `imageantialias` is generally available after PHP 7.2.0. For versions before that, this check is needed.

### DIFF
--- a/src/Gd/Drawer.php
+++ b/src/Gd/Drawer.php
@@ -140,7 +140,9 @@ final class Drawer implements DrawerInterface
         if ($thickness === 0 && !$fill) {
             return $this;
         }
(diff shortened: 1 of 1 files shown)


ahukkanen:
Could also be a good idea to add one CI environment which uses a PHP version below 7.2.0 and has been compiled without GD, in which case this function could be unavailable.

ausi:
For clarification to other reviewers: PHP can be configured to use the bundled GD version (`--with-gd`) or to use a custom one (`--with-gd=DIR`). If the bundled version is used the `GD_BUNDLED` constant is set to `1`.

Before PHP 7.2 the `imageantialias()` function was only available if the bundled GD version was used.

PR #714, opened 2019-05-28
feature - more information in RuntimeException

Raneomik:
porposal of error_get_last usage for having more verbose error messages in RuntimeException

see [>this issue#1167<](https://github.com/liip/LiipImagineBundle/issues/1167) encountered while using LiipImagineBundle in a SF4 app

### DIFF
--- a/src/Exception/RuntimeException.php
+++ b/src/Exception/RuntimeException.php
@@ -11,9 +11,21 @@
 
 namespace Imagine\Exception;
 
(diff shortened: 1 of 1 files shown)


mlocati:
IMHO this is a bit a hack: we should build a better error message when throwing the exception, not in its constructor

mlocati:
Closing because of missing feedback

PR #723, opened 2019-09-16
Fix converting of gif to jpg with imagick adapter

alexander-schranz:
Flatten does accidently reset the format to gif and so it need to be set after `$this->flatten`

### DIFF
--- a/src/Imagick/Image.php
+++ b/src/Imagick/Image.php
@@ -404,10 +404,6 @@ final class Image extends AbstractImage
      */
     private function prepareOutput(array $options, $path = null)
     {
-        if (isset($options['format'])) {
-            $this->imagick->setImageFormat($options['format']);
-        }
-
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
 sudo: true
+dist: trusty # needed for php 5.5 and smaller
```
alexander-schranz: this seems to fix 5.4 and 5.5 but 5.3 still fails @mlocati any idea?

mlocati: I found the reason - fixed by https://github.com/avalanche123/Imagine/pull/744

alexander-schranz: @mlocati Did rebase the PR

mlocati:
I'm going to address this issue in #746

alexander-schranz:
@mlocati thank you!

PR #729, opened 2019-12-04
Improve detection of unsupported Exit Metadata Reader

mlocati:


mlocati:
PS: I've just released [version 1.2.3](https://github.com/avalanche123/Imagine/releases/tag/1.2.3) which includes this patch.

PR #731, opened 2019-12-17
Use PHP 7.4 instead of a snapshot on Travis

andreybolonin:


mlocati:
I'd also remove the `- name: Test with PHP 7.4 (all drivers)` line right under the `allow_failures:` section

mlocati:
:+1:

Issue #732, opened 2019-12-19
GMagick No encode delegate for this image format


ausi:
### Issue description

On my system it is not possible to save a GMagick image to a file without a file extension. The Exception `No encode delegate for this image format (jpeg)` gets thrown.

### What's the PHP version you are using?

`7.3.13-1+ubuntu16.04.1+deb.sury.org+1`

### What's the imaging library you are using [gd/imagick/gmagick/any]?

GMagick

### What's the imaging library configuration 

```
gmagick version => 2.0.5RC1
GraphicsMagick version => GraphicsMagick 1.4 snapshot-20180922 Q16 http://www.GraphicsMagick.org/
```

### Minimal PHP code to reproduce the error: 

```php
$gmagick = new \Gmagick();
$gmagick->newimage(100, 100, (new \GmagickPixel('#ffffff'))->getcolor(false));
$gmagick->setimageformat('jpeg');
$gmagick->writeimage(__DIR__.'/testimage');
```

### How to fix

The [GraphicsMagick documentation](http://www.graphicsmagick.org/wand/magick_wand.html#magicksetimageformat) shows the image format in uppercase (`"GIF"`) and if I replace the `'jpeg'` with `'JPEG'` in the test code above this fixes the issue for me.

I think we should replace https://github.com/avalanche123/Imagine/blob/3db069bd870f45b5305e7d6bd4e08d45ebb22cef/src/Gmagick/Image.php#L433 with:

```php
$this->gmagick->setimageformat(strtoupper($options['format']));
```

ausi:
@mlocati should I provide a pull request for the fix?

mlocati:
@ausi that'd be great

ausi:
Pull request: #733

PR #733, opened 2020-01-15
Use uppercase format string for GMagick

ausi:
This prevents the GMagick error `No encode delegate for this image format`

See #732

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ script:
             composer --no-interaction codestyle -- --no-interaction --dry-run --diff -- .
             ;;
         *)
-            composer --no-interaction test -- --exclude-group "${PHPUNIT_EXCLUDED_GROUPS}"
+            composer --no-interaction test -- --exclude-group "${PHPUNIT_EXCLUDED_GROUPS}" --filter=testSaveWithoutFileExtension
             ;;
     esac
   - if test -n "${AUTOUPDATE_API:-}"; then ./.travis/autoupdate-api.sh; fi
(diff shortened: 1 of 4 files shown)


ausi:
The failing test on Travis doesn’t seem to be related to this pull request.

ausi:
@mlocati it would be great if this could be merged ☺️

leofeyer:
I second this, because we are also having issues with this bug (see https://github.com/contao/image/issues/67).

mlocati:
We don't need to second a change, we need a test case that shows the problem.

The correct way to submit a pull request should be:

1. create a pull request with a commit that makes test fail because of a bug
2. wait for TravisCI tests to fail
3. add to the same pull request a second commit that fixes the bug
4. tests should now succeed

I just created #737 that tests saving images with different cases of the `format` parameter (see [here](https://github.com/avalanche123/Imagine/pull/737/files)), and the tests are passing for all the 3 drivers GD, GMagick and Imagick (see [here](https://travis-ci.org/avalanche123/Imagine/builds/660165501)).

So, how can we replicate the problem that this pull request is addressing?

leofeyer:
@mlocati No hard feelings, but I don't understand your reaction. @ausi has explained the issue very detailed in #732 and you have explicitly asked him to create a PR:



In the original issue, @ausi states that the problem is not reproducible on every system. Apparently, it does not occur on Travis. 🤷‍♂

It still does occur on @ausi's system and on our servers. And the GraphicsMagick documentation, which @ausi has also linked in #732, says that the "magick string" is an uppercase string. Isn't that reason enough to merge such a little change?

I understand that unit tests are important to prevent regression bugs. But IMHO it is not worth the effort to fiddle with Travis until we have a setup in which we can reproduce this, just so we can add a simple `strtoupper()` in a place where – at least according to the docs – it should have been an upper-case string from the beginning.

Again, no hard feelings. 🙈

mlocati:
> No hard feelings, but I don't understand your reaction.

I didn't meant to be rude, of course. Maybe using one of those great emoticons would have made my reaction much more polite :wink:

In the [documentation](http://www.graphicsmagick.org/wand/magick_wand.html#magicksetimageformat) we don't have a clear note about upper/lower case, only an example with `'GIF'`.

> Isn't that reason enough to merge such a little change?

Well, are you sure that other systems would not break if we use `JPEG` instead of `jpeg`?

I'm going to check the GraphicsMagic and GMagick source code. Maybe something changed betweem the GraphicsMagick version 1.3.23 (used in TravisCI) and 1.4 (as reported in #732)

leofeyer:
> Well, are you sure that other systems would not break if we use `JPEG` instead of `jpeg`?

Well, no. 😞 Good point.

mlocati:
The `setimageformat` function of the GMagick PHP extension is a [bare wrapper](http://svn.php.net/viewvc/pecl/gmagick/trunk/gmagick_methods.c?view=markup#l1897) around the [`MagickSetImageFormat`](http://hg.code.sf.net/p/graphicsmagick/code/file/872beb168a36/wand/magick_wand.c#l8611) function of the GraphicsMagick library.
This `MagickSetImageFormat` function simply set the `magick` property of the GraphicsMagick `Image` object.
That `magick` property is then [used](http://hg.code.sf.net/p/graphicsmagick/code/file/872beb168a36/magick/constitute.c#l2293) to determine the image info, using the [`GetMagickInfo`](http://hg.code.sf.net/p/graphicsmagick/code/file/872beb168a36/magick/magick.c#l411) function. That function calls [`GetMagickInfoEntryLocked`](http://hg.code.sf.net/p/graphicsmagick/code/file/872beb168a36/magick/magick.c#l374), which uses [`LocaleCompare`](http://hg.code.sf.net/p/graphicsmagick/code/file/872beb168a36/magick/utility.c#l3351) to check the format, which performs a case-insensitive match.
So, I really don't understand why or when the issue may occur...

ausi:
@mlocati in d8bc9bffea7281bd1f7bc43d5805fea2521c4496 I created a reproduction of the issue, see [Travis build 1436](https://travis-ci.org/avalanche123/Imagine/builds/660749461).

In f71dff28a95427a89ba34fcfc0da966fa378d5f7 the issue gets fixed, see [Travis build 1438](https://travis-ci.org/avalanche123/Imagine/builds/660752487).  
_(The build itself fails but the tests pass)._

mlocati:
Superseded by #750

PR #738, opened 2020-03-09
Try to reproduce the 'no encode delegate' in the unit tests

leofeyer:
See #732 and #733.


leofeyer:
I think I have managed to reproduce the issue on Travis:



https://travis-ci.org/avalanche123/Imagine/jobs/660247300#L1681-L1685

Next I am going to commit the fix and see if the builds pass.

mlocati:
> I think I have managed to reproduce the issue on Travis:

Well, also the `WEBP` test case fails, not only the `webp` one...

leofeyer:
Correct. The fix did not fix the issue. 😞

leofeyer:
@ausi told me that he will have a look at this again tonight.

PR #740, opened 2020-09-13
Fix compatibility with PHP 8.0

ausi:
In PHP 8.0 the GD resources got replaced with `GdImage` objects, see https://github.com/php/php-src/commit/8aad466c42b5feabc7565856eb006c326f35382f

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
     esac
   - php --ri curl || true
   - ifconfig || true
-  - composer --no-interaction install --ignore-platform-reqs
+  - composer --no-interaction install
 
 script:
   - |
@@ -110,6 +110,9 @@ matrix:
     -
       name: Test with PHP nightly (all drivers)
       php: nightly
+      before_install:
+        - phpenv config-rm xdebug.ini || true
+        - composer --no-interaction config platform.php 7.4.0
       env:
         - IMAGINE_DRIVER: imagick
     -
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
         *)
             sudo apt-get update -y || true
             sudo apt-get remove -y imagemagick libmagickcore-dev libmagickwand-dev
-            sudo apt-get install -y ca-certificates libtiff-dev libjpeg-dev libdjvulibre-dev libwmf-dev pkg-config liblcms2-dev
+            sudo apt-get install -y ca-certificates libtiff-dev libjpeg-dev libdjvulibre-dev libwmf-dev pkg-config liblcms2-dev autoconf
             PHPUNIT_EXCLUDED_GROUPS=always-skipped
             if test "${IMAGINE_DRIVER:-}" = 'imagick' -o "${IMAGINE_DRIVER:-}" = 'all'; then ./.travis/imagick.sh; else PHPUNIT_EXCLUDED_GROUPS="${PHPUNIT_EXCLUDED_GROUPS},imagick"; fi
             if test "${IMAGINE_DRIVER:-}" = 'gmagick' -o "${IMAGINE_DRIVER:-}" = 'all'; then ./.travis/gmagick.sh; else PHPUNIT_EXCLUDED_GROUPS="${PHPUNIT_EXCLUDED_GROUPS},gmagick"; fi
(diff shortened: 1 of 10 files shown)


ausi:
It looks like the `nightly` build doesn’t run through https://travis-ci.org/github/avalanche123/Imagine/jobs/726730923  
I don’t know how to fix this.

mlocati:
I [fixed](https://github.com/avalanche123/Imagine/pull/740/commits/5efaeb08515fa2bfbd6d4b5156073eeede6f6929) the compilation of the `imagick` PHP extension, but it seems that `gmagick` extension is not ready for PHP 8 (see https://pecl.php.net/package/gmagick and http://svn.php.net/viewvc/pecl/gmagick )

mlocati:
PS: for `gmagick` I submitted [this bug report](https://bugs.php.net/bug.php?id=80106).

.travis.yml
```diff
@@ -111,7 +111,7 @@ matrix:
       name: Test with PHP nightly (all drivers)
       php: nightly
       env:
-        - IMAGINE_DRIVER: all
+        - IMAGINE_DRIVER: imagick
```
ausi, line 114: This should be reverted once the GMagick issue has been resolved, <https://bugs.php.net/bug.php?id=80106>

.travis.yml
```diff
@@ -33,7 +33,7 @@ install:
     esac
   - php --ri curl || true
   - ifconfig || true
-  - composer --no-interaction install
+  - composer --no-interaction install --ignore-platform-reqs
```
ausi, line 36: This should be reverted once the dev dependencies are compatible with PHP 8.0.

ausi:
Besides the GMagick error and the incompatible dev-dependencies PHP 8.0 compatibility should be OK with this pull request I think.

.travis.yml
mlocati: Instead of `--ignore-platgorm-reqs` we could tell Composer to assume that the current PHP version is 7.4 instead of 8 (see https://github.com/mlocati/ip-lib/blob/1.12.0/.travis.yml#L101 for an example)

ausi: Done in b21c12305cbd114b35b436486d83dabb152a6fc1

mlocati (approved):

ausi:
It looks like there is not much activity in GMagick (see https://github.com/vitoc/gmagick/pull/48 ). 

@mlocati should we merge this as it is now to have at least GD and IMagick support for PHP 8.0?

src/Gd/Layers.php
```diff
@@ -50,7 +50,7 @@ class Layers extends AbstractLayers
      */
     public function __construct(Image $image, PaletteInterface $palette, $resource, $initialOffset = 0)
     {
-        if (!is_resource($resource)) {
+        if (!$resource instanceof \GdImage && !is_resource($resource)) {
```
derrabus, line 53: The `@param` annotation for `$resource` should be adjusted as well:

```
@param resource|\GdImage $resource
```

src/Gd/Imagine.php
```diff
@@ -91,7 +91,7 @@ public function open($path)
 
         $resource = $this->createImageFromString($data);
```
derrabus, line 92: The doc block of `createImageFromString` does not reflect the new type yet.

```
@return resource|\GdImage|false
```

derrabus (changes requested):
Thank you for fixing this issue. ❤️ 

I'm using this library in an application that I'd like to test with php 8. Merging at least the fix for the GD incompatibilities would really help me out.

src/Gd/Layers.php
ausi: Fixed in e424a69a354dd46a57467a841128bad10bc249e2

src/Gd/Imagine.php
ausi: Fixed in e424a69a354dd46a57467a841128bad10bc249e2

ausi:
> Merging at least the fix for the GD incompatibilities would really help me out.

That’s pretty much everything this PR does. There don’t seem to be any other incompatibilities with PHP 8.0.

The planned release date for PHP 8.0 is less than a month away, so I think it would be a good time to merge this PR ☺️

PR #741, opened 2020-09-25
Read Exif data from local files directly

jorrit:
Reading the file data into memory is not needed for local files. This reduces memory consumption in ExifMetadataReader.


mlocati:
Thanks!

PR #747, opened 2020-11-03
Fix GMagick No encode delegate

mlocati:


ausi:
If I remember correctly from #733 the error didn’t always show up. It seems that this issue only occurs before the first successful save operation for this format happened. Maybe adding `@runInSeparateProcess` to the test method can help to reveal the issue.

mlocati:
> Maybe adding `@runInSeparateProcess` to the test method can help to reveal the issue.

Yep, it worked! And also broke PHP with a segfault 🤣

tests/tests/Image/AbstractImageTest.php
```diff
@@ -597,6 +597,26 @@ public function testInOutResult($file, $in, $out)
         $thumb->save($filename);
     }
 
+    /**
+     * @dataProvider inOutResultProvider
+     * @runInSeparateProcess
```
ausi, line 602:
```suggestion
     * @runInSeparateProcess
     * @preserveGlobalState disabled
```
Maybe this helps with the PHP segfault.

mlocati:
Any suggestion to fix those `Segmentation fault (core dumped)` errors is very welcome...

mlocati:
I tried with https://github.com/avalanche123/Imagine/pull/747/commits/c7dfccd2314536fa448eed68da9b2d97b47f62c5#diff-4c0aab9c8ed7ce313e9c5822412ee5e575471f32232b34724e15b2072771b2c3 and in this case there's no segfault, so the issue is with PHPUnit (see https://travis-ci.com/github/avalanche123/Imagine/builds/198035883)

ausi:
Maybe we should fix the issue without having it captured by the CI for now?

mlocati:
I guess it's the only solution... do you mind submitting a new pull request with only the fix?

ausi:
Done: #750

PR #756, opened 2020-12-28
Convert unsupported colour space to RGB on imagick

### LINKED ISSUE #755: Add support for HEIC format
Since iphone 11 and up stores images in HEIC format, it would be really nice to support it.

ImageMagick supports HEIC images (needs to be compiled with libheif and libde265).

Imagine fails to open a HEIC file with the following error:

    Only RGB and CMYK colorspace are currently supported

A sample image taken with my iphone 11 says the color space is `YCbCr`. Not sure how difficult it would be to add support for this colorspace and if any other modifications would be required to have HEIC image processing working.

jleckie:
This removes the limitation that only CMYK and RGB are supported on imagemagick.  To avoid having to add colourspace support for the dozens of different systems supported by imagemagick, it simply converts other colour spaces to SRGB for simplicity when dealing with colour in imagine.  

This will cause problems, if you wanted to handle colour profiles before the conversion, but it seems to work well enough for my use cases.  

This has the added benefit of allowing HEIC files to be supported and should resolve #755.  HEIC files often are in either [dci-p3](https://en.wikipedia.org/wiki/DCI-P3) (Apple Phones) or  [CIELAB](https://en.wikipedia.org/wiki/CIELAB_color_space) (Nokia?), or YCbCr even.  

In order to get HEIC support I needed to build my own imagemagick and imagick libraries from source.  There is a few guides around online for this, obviously you will need to customise for your own environment.
https://gist.github.com/caingougou/e2ebe70057270c208d57ed13ec186b7e


ausi:
I integrated part of this in my HEIC/AVIF pull request #759.

Do you have any test images that use something different than `Imagick::COLORSPACE_YCBCR` (that are not `RGB` or `CMYK`)?

I think it would be more safe to only add support for `YCBCR` for now.

jleckie:
Closing my PR as it has been superseded by @ausi

PR #758, opened 2021-01-16
Add test agains php 8

alexander-schranz:
Should we also migrate the tests to github actions?


mlocati:
Thank you.
BTW I'm going to rewrite the tests.

Issue #742, opened 2020-10-28
Add support for AVIF format


jonnitto:
Currently, these files formats are supported: jpg, jpeg, gif, png, wbmp, xbm, webp, and bmp

But there is already the next generation of image formats here: AVIF. [You can read more about it here](https://reachlightspeed.com/blog/using-the-new-high-performance-avif-image-format-on-the-web-today/)

With this option, we could even serve faster websites with PHP and this library 🚀

ausi:
👍 AVIF support would be great!

But I think we have to wait until the underlying libraries that Imagine uses add support the AVIF format:

- GD: https://github.com/libgd/libgd/issues/557
- IMagick: https://github.com/Imagick/imagick/issues/356
- GMagick: I could’t find any issue in https://github.com/vitoc/gmagick for AVIF support

SoftCreatR:
Imagick already supports AVIF. It just depends on the installed ImageMagick version and the installed/compiled delegates:

![image](https://user-images.githubusercontent.com/81188/97717405-0b931480-1ac5-11eb-9435-9f153f2d2143.png)

Therefore i guess, that GMagick also supports AVIF already.

As for the ImageMagick installation, you may check this out: https://github.com/SoftCreatR/imei

jonnitto:
@ausi Could create an issue for GMagick?. I already created also one for [imagine/imagine](https://github.com/avalanche123/Imagine/issues/742) package doesn't support `avif` yet.

SoftCreatR:
@jonnitto The problem is, that GMagick is just a wrapper around GraphicsMagick and according to [this](http://www.graphicsmagick.org/formats.html) list, neither HEIC nor AVIF is supported, yet.

Renkas:
But why not open a issue in GraphicsMagick's tracker then?

markusguenther:
> But why not open a issue in GraphicsMagick's tracker then?

Found this repository but not sure if they are using github somewhere?
https://sourceforge.net/p/graphicsmagick/code/ci/default/tree/

ausi:
It seems that there is interest in supporting AVIF from @bobfriesenhahn (the maintainer of GraphicsMagick): https://sourceforge.net/p/graphicsmagick/discussion/250738/thread/3082583888/

The issue tracker for feature requests is hosted here I think: https://sourceforge.net/p/graphicsmagick/feature-requests/

ausi:
Pull request for AVIF support in Imagick driver: #759

[...]

PR #759, opened 2021-02-13
Support AVIF and HEIC in Imagick driver

ausi:
This adds support for AVIF and HEIC in the Imagick driver as suggested in #742 and #755.

~~This is still a work in progress. Saving files already works fine, but reading images files currently fails because imagemagick (incorrectly?) reports `Imagick::COLORSPACE_YCBCR` as the colorspace even though it is RGB. Just ignoring this doesn’t seem to work because the color profile doesn’t get interpreted correctly then (tested with an iPhone HEIC image).~~


ausi:
I was able to fix the colorspace issue by the suggested fix from @jleckie in #756. Now the color profile gets applied correctly and the colors stay the same with my test files.

ausi:
The failing tests do not seem to be related to this pull request.

laxity7:
So what about this MR? Obviously, the solution is working, the problem is outside of this MR. 

Can you accept it? Doesn't work with heic without it

mlocati:
@ausi Do you have a working environment for imagick + HEIC?
I'm trying to fix the tests: they now use Docker images (built by https://github.com/php-imagine/docker-builder/blob/main/.github/workflows/create-images.yml and https://github.com/php-imagine/docker-builder/blob/main/docker/installer.sh)
In the imagick tests we have errors like [this one](https://github.com/php-imagine/Imagine/runs/3807085250?check_suite_focus=true#step:7:39):

```
Imagine\Test\Imagick\ImageTest::testSaveCompressionQuality with data set #5 ('heic', array(0), array(100))
Imagine\Exception\RuntimeException: Save operation failed

src/Imagick/Image.php:343
tests/tests/Image/AbstractImageTest.php:1031
tests/tests/Imagick/ImageTest.php:179

Caused by
ImagickException: no encode delegate for this image format `HEIC' @ error/constitute.c/WriteImage/1272

src/Imagick/Image.php:341
tests/tests/Image/AbstractImageTest.php:1031
tests/tests/Imagick/ImageTest.php:179
```

ausi:
Do we need to add the `--with-heic` and other flags to https://github.com/php-imagine/docker-builder/blob/f0e428b7ec295aef48ae674711cb4cbf81ad22e3/docker/installer.sh#L235 ?

mlocati:
Nope: it's detected automatically: see https://github.com/php-imagine/docker-builder/runs/3806674769?check_suite_focus=true#step:11:1144

I'll try to see of installing some additional libraries fixes the issue (the script you linked above is very helpful)

mlocati:
I understood what's missing: libheif uses libde265 to read heic files, and libx265 to write them: I only installed libde265

mlocati:
See https://github.com/php-imagine/docker-builder/pull/2/commits/24e54b8b28f6b184a25fba1541b128a3991ca4be

laxity7:
When will the release be?
The master does not yet have these changes, but it is very necessary.

And thanks for being active

PR #760, opened 2021-03-22
remove final from drivers

mynameisbogdan:
Hello,

When coupled with `liip/imagine-bundle`, these classes can be set as lazy in symfony since they aren't necessary until resolving is needed.

### DIFF
--- a/src/Gd/Imagine.php
+++ b/src/Gd/Imagine.php
@@ -26,8 +26,10 @@ use Imagine\Utils\ErrorHandling;
 
 /**
  * Imagine implementation using the GD library.
+ *
+ * @final
  */
-final class Imagine extends AbstractImagine
+class Imagine extends AbstractImagine
 {
     /**
      * Initialize the class.
--- a/src/Gmagick/Imagine.php
+++ b/src/Gmagick/Imagine.php
@@ -27,6 +27,8 @@ use Imagine\Image\Palette\RGB;
 
 /**
  * Imagine implementation using the Gmagick PHP extension.
+ *
+ * @final
  */
 class Imagine extends AbstractImagine
 {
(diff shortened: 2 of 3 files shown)


ausi:
Shouldn’t it be possible to load the service lazily by specifying the interface?

`lazy: Imagine\Image\ImagineInterface`

mynameisbogdan:
Hello @ausi,

According to the (docs)[https://symfony.com/doc/current/service_container/lazy_services.html] seems to not be possible.

As well a warning: ***Lazy services do not support final classes.***

ausi:
I think this was added in Symfony version 4.2, see https://github.com/symfony/symfony/pull/27697

mynameisbogdan:
I was looking at `Definition::public function setLazy(bool $lazy)` and I wasn't aware of proxied interfaces.

You're correct, ```$definition->addTag('proxy', ['interface' => \Imagine\Image\ImagineInterface::class]);``` seems to do the trick.

Personally I would prefer a non-final class, but the final decision is yours. A proxied interface would do for the time being. 

Thank you for taking the time to discuss this changes.

ausi:
> Personally I would prefer a non-final class

I have no preference either way.  
But I think we should make it consistent because currently `Gd\Imagine` and `Imagick\Imagine` are `final` but `Gmagick\Imagine` is not.

mynameisbogdan:
> But I think we should make it consistent because currently `Gd\Imagine` and `Imagick\Imagine` are `final` but `Gmagick\Imagine` is not.

Yes, I noticed that as well. This PR would make them consistent.

mlocati:
Yep, it doesn't seem necessary for lazy loading, but letting people overriding classes may be helpful.

PR #765, opened 2021-06-12
Support JXL in Imagick driver

ausi:
This adds support for JPEG-XL in the Imagick driver.

@mlocati As this is similar to #759 (AVIF and HEIC) should I add these changes to the other pull request and close this one?


ausi:
@mlocati should I rebase this pull request?

src/Imagick/Image.php
```diff
@@ -848,6 +848,23 @@ private function applyImageOptions(\Imagick $image, array $options, $path)
...
+                    $image->setimagecompressionquality(100);
+                    $image->setcompressionquality(100);
+                } else {
+                    if (!isset($options[$format . '_quality'])) {
+                        if (isset($options['quality'])) {
+                            $options[$format . '_quality'] = $options['quality'];
+                        }
+                    }
+                    if (isset($options[$format . '_quality'])) {
+                        $options[$format . '_quality'] = max(9, min(99, $options[$format . '_quality']));
```
mlocati, line 862: Is the `9` here correct?

mlocati:
> @mlocati should I rebase this pull request?

@ausi I've just did that :wink:

src/Imagick/Image.php
ausi: I think so, yes. But I can’t seem to remember why now. Give me a second... ☺️

mlocati: > I think so, yes. But I can’t seem to remember why now. Give me a second... ☺️

I understand... sorry for the delay in reviewing this PR...

ausi: Just tested it manually and qualities from 0 to 8 all create the exact same image with a size about five times larger than quality 9.

I think we should merge it as it is now and probably create a new ticket to check if this is a bug that needs to be fixed in ImageMagick.

PR #768, opened 2021-07-21
Fix PHP 8.1 deprecations

ausi:
Fixes the PHP 8.1 deprecations of:

1. https://wiki.php.net/rfc/internal_method_return_types
2. https://wiki.php.net/rfc/deprecations_php_8_1#num_points_parameter_of_image_open_filled_polygon

We should probably wait with merging this until an RC version of PHP 8.1 was released.

### DIFF
--- a/src/Gd/Drawer.php
+++ b/src/Gd/Drawer.php
@@ -331,7 +331,11 @@ final class Drawer implements DrawerInterface
             throw new RuntimeException('Draw polygon operation failed');
         }
 
-        if (false === $callback($this->resource, $points, count($coordinates), $this->getColor($color))) {
+        if (false === (
+            PHP_VERSION_ID < 80000
+                ? $callback($this->resource, $points, count($coordinates), $this->getColor($color))
+                : $callback($this->resource, $points, $this->getColor($color))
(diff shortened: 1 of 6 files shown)


ausi:
PHP version 8.1.0-RC1 was released a month ago so I think this pull request is ready to merge.

HVSoftware:
@ausi I see this fix is merged into the develop branch. Only this fix is not applies into the master branch. For this reason I cannot update to PHP 8.1 when using the master branch. Using the develop branch is not an option within my company. Can this fix also be applied to the master branch?

ausi:
I think you have to wait until the next release. As far as I know the following issues have to be fixed before the version 1.3.0 can be released: https://github.com/php-imagine/Imagine/milestone/1

mlocati:
Yep, we have a few issues that still need to be fixed. I've spent an insane amount of time trying to fix them, but no luck.

I'll try to find again some spare time to work on them (of course, any help is welcome)

HVSoftware:
Clear. I will take a look at one of those issues. I'm not saying I can fix them. Only it would be nice if it can.

PR #770, opened 2021-09-30
Move PHP Coding Style Check to GitHub Action

mlocati:


mlocati:
@avalanche123 GitHub Actions are disabled for this repository? It seems that the action added here isn't executed...

PR #792, opened 2021-10-08
Fix Imagine\Test\Imagick\EffectsTest::testGrayscale

ausi:
_Currently draft status to see where the build fails_


mlocati:
@ausi In #795 I added a way to save the temporary files as artifacts. If you need to see why a test fails, you can [uncomment this line](https://github.com/php-imagine/Imagine/blob/ede7542da670ff2d434525f7b37ea304b3cadf37/.github/workflows/phpunit.yml#L22) and download the relevant artifacts to better understand what went wrong.

PR #802, opened 2021-10-15
Add driver info

mlocati:
This PR lets users know in advance the available drivers, as well as checking their limitations.


vlakoff:
Refs #835

PR #816, opened 2022-02-11
Fix CMYK palette issue

xKZL:
JPEG images saved with CMYK palette gets color distorded when processed with Gd.


xKZL:
I have implemented a fix for this issue, can someone double check that ?
Thanks..

PR #819, opened 2022-02-15
Fix deprecations in MetadataBag

xKZL:

### DIFF
--- a/src/Image/Metadata/MetadataBag.php
+++ b/src/Image/Metadata/MetadataBag.php
@@ -47,8 +47,7 @@ class MetadataBag implements \ArrayAccess, \IteratorAggregate, \Countable
      *
      * @see \Countable::count()
      */
-    #[\ReturnTypeWillChange]
-    public function count()
+    public function count(): int
     {
         return count($this->data);
     }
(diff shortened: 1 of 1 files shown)


mlocati:
Isn't `#[\ReturnTypeWillChange]` enough?

xKZL:
I get dozen of warnings right now with using PHP 8.1.

In my understanding, MetadataBag is not the player in charge of the return type, but "ArrayAccess".
So it should just be updated accordingly. Unless I missed something.

mlocati:
> I get dozen of warnings right now with using PHP 8.1.

Do you have the `#[\ReturnTypeWillChange]` attribute in your code?

ausi:
> I get dozen of warnings right now with using PHP 8.1.

I was not able to reproduce that. Are you using the `dev-develop` version or `1.2.4`?

mbabker:
You need a combination of `#[\ReturnTypeWillChange]` to silence the PHP engine deprecations and a `@return` annotation to silence the deprecations coming from Symfony's debug class loader.

Adding the native return types is a B/C break in an open class, and as is breaks PHP 7 support with the mixed type.

PR #832, opened 2022-09-01
imagick: add reset image page after resize

4rthem:
Chaining `resize` then `crop` produced invalid output.


ausi:
How can I reproduce the issue?

Can you add a unit test that tests this?

PR #835, opened 2022-11-15
Fix invalid format MIME type definition

xdanik:
I have discovered that $image->show() was setting invalid Content-Type header. "jpg" in case of "jpg" format instead of "image/jpg".

There seems to be an invalid argument order in the constructor call to Imagine\Image\Format.
The constructor signature is:
`private function __construct($id, $mimeType, $canonicalFileExtension, $alternativeIDs = array())`

While callers were clearly setting MIME type into 3. argument instead of second one.

### DIFF
--- a/src/Image/Format.php
+++ b/src/Image/Format.php
@@ -154,7 +154,7 @@ class Format
             case static::ID_JPEG:
                 return new static($formatID, 'image/jpeg', 'jpg', array('jpg', 'pjpeg', 'jfif'));
             case static::ID_WBMP:
-                return new static($formatID, 'vnd.wap.wbmp', 'jpg');
+                return new static($formatID, 'image/vnd.wap.wbmp', $formatID);
             default:
                 return new static($formatID, "image/{$formatID}", $formatID);
         }
--- a/src/Image/Format.php
+++ b/src/Image/Format.php
@@ -152,11 +152,11 @@ class Format
     {
         switch ($formatID) {
             case static::ID_JPEG:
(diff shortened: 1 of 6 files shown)


src/Image/Format.php
```diff
@@ -152,11 +152,11 @@ protected static function create($formatID)
     {
         switch ($formatID) {
             case static::ID_JPEG:
-                return new static($formatID, 'jpg', 'image/jpeg', array('jpg', 'pjpeg', 'jfif'));
+                return new static($formatID, 'image/jpeg', 'jpg', array('jpg', 'pjpeg', 'jfif'));
             case static::ID_WBMP:
-                return new static($formatID, 'jpg', 'vnd.wap.wbmp');
+                return new static($formatID, 'vnd.wap.wbmp', 'jpg');
```
ausi, line 157: Shouldn’t the extension be `wbmp` here?
And the correct mime type starts with `image/` I think.
```suggestion
                return new static($formatID, 'image/vnd.wap.wbmp', $formatID);
```

xdanik: Yes, I guess it should. I have just swapped the arguments order. Committed now.

ausi (approved):
Adding a *tests/tests/Image/FormatTest.php* would make sense too I think ☺️

xdanik:
I have added simple unit test that asserts expected MIME types for a few basic image formats.

ausi (approved):

vlakoff:
Nice catch! I noticed you also fixed the file extension for WBMP.

PR #843, opened 2023-01-25
PHP 8 compatibility.

glodov:
Fix for
The error is thrown in the HumHub project:

PHP Deprecated Warning 'yii\\base\\ErrorException' with message 'Implicit conversion from float-string "42.95454545454545" to int loses precision' 

in /www/wwwroot/yta.yaro.info/protected/vendor/imagine/imagine/src/Imagick/Image.php:164

Stack trace:
#0 [internal function]: yii\\base\\ErrorHandler->handleError() #1 /www/wwwroot/yta.yaro.info/protected/vendor/imagine/imagine/src/Imagick/Image.php(164): Imagick->cropImage() #2 /www/wwwroot/yta.yaro.info/protected/humhub/libs/ProfileImage.p...",

### DIFF
--- a/src/Image/Point.php
+++ b/src/Image/Point.php
@@ -38,6 +38,12 @@ final class Point implements PointInterface
      */
     public function __construct($x, $y)
     {
+        if (!\is_int($x)) {
+            $x = (int) round($x);
+        }
+        if (!\is_int($y)) {
+            $y = (int) round($y);
+        }
         if ($x < 0 || $y < 0) {
             throw new InvalidArgumentException(sprintf('A coordinate cannot be positioned outside of a bounding box (x: %s, y: %s given)', $x, $y));
         }
--- a/src/Imagick/Image.php
+++ b/src/Imagick/Image.php
@@ -161,7 +161,7 @@ final class Image extends AbstractImage implements InfoProvider
                 }
                 $this->imagick = $this->imagick->deconstructImages();
             } else {
-                $this->imagick->cropImage((int)$size->getWidth(), (int)$size->getHeight(), (int)$start->getX(), (int)$start->getY());
+                $this->imagick->cropImage($size->getWidth(), $size->getHeight(), $start->getX(), $start->getY());
                 // Reset canvas for gif format
                 $this->imagick->setImagePage(0, 0, 0, 0);
             }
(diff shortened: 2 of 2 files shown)


ausi:
How can this error be reproduced?
`$size` and `$start` should always contain integer values.

glodov:
> How can this error be reproduced? `$size` and `$start` should always contain integer values.

I just had it with PHP version 8.
It seems some functions send numbers as string values.

ausi:
> I just had it with PHP version 8.
> It seems some functions send numbers as string values.

I was not able to reproduce that.
How does your code in *humhub/libs/ProfileImage.p...* look like? Maybe the issue has its cause there.

ausi:
> You can also try to pass a string as a number '8.123' or just float the number '8.123'.

I think I see the problem now:

```php
$box = new Box('123.45', '123.45');
$point = new Point('123.45', '123.45');

var_dump($box->getWidth()); // int(123)
var_dump($point->getX()); // string(6) "123.45"
```

So I think this needs to be fixed in the `Point` class similar to how it is handled in the `Box` class: https://github.com/php-imagine/Imagine/blob/c10ea7f80e5b93473e2390515adc3057b85e26a2/src/Image/Box.php#L41-L46

glodov:
@mlocati can you apply the pull request?

ausi (approved):

src/Image/Point.php
```diff
@@ -38,6 +38,12 @@ final class Point implements PointInterface
      */
     public function __construct($x, $y)
     {
+        if (!\is_int($x)) {
```
mlocati, line 41: If we pass wrong arguments, they will be [silently converted to 0](https://3v4l.org/FKFSe).

I'd instead do something like

```php
        if (!is_numeric($x)) {
            throw new InvalidArgumentException('x must be numeric');
        }
        $x = (int) $x;
        if (!is_numeric($y)) {
            throw new InvalidArgumentException('y must be numeric');
        }
        $y = (int) $y;
```

ausi: I think for better consistency, this should be handled the same way as `Box` class does: https://github.com/php-imagine/Imagine/blob/c10ea7f80e5b93473e2390515adc3057b85e26a2/src/Image/Box.php#L41-L46

ausi: Ah, I didn’t think of the `0` case. 
So +1 for you latest code suggestion 👍

PR #847, opened 2023-06-05
remove warning, Deprecated: Implicit conversion from float to int loses precision in PHP 8.2

anovsiradj:
remove deprecated message from php8.2

edit: i guest this PR related to #843 😳

`Deprecated: Implicit conversion from float 345.5 to int loses precision in ...\vendor\imagine\imagine\src\Gd\Image.php on line 191`

on https://github.com/php-imagine/Imagine/blob/6d5b83a403d17236ba3b7e52e3c2caf4e937d0f0/src/Gd/Image.php#L191
`imagecopymerge()` need `dst_x` and `dst_y` to be int.

but `$start->getX()` and `$start->getY()` could return float.

### DIFF
--- a/src/Image/Point.php
+++ b/src/Image/Point.php
@@ -42,8 +42,8 @@ final class Point implements PointInterface
             throw new InvalidArgumentException(sprintf('A coordinate cannot be positioned outside of a bounding box (x: %s, y: %s given)', $x, $y));
         }
 
-        $this->x = $x;
-        $this->y = $y;
+        $this->x = ceil($x);
+        $this->y = ceil($y);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


mlocati:
Why ceil instead of round?

src/Image/Point.php
```diff
@@ -42,8 +42,8 @@ public function __construct($x, $y)
             throw new InvalidArgumentException(sprintf('A coordinate cannot be positioned outside of a bounding box (x: %s, y: %s given)', $x, $y));
         }
 
-        $this->x = $x;
-        $this->y = $y;
+        $this->x = ceil($x);
+        $this->y = ceil($y);
```
ausi, line 46: I think the solution outlined in https://github.com/php-imagine/Imagine/pull/843#discussion_r1143490195 would be the correct fix here.

anovsiradj:
hello, i have revert previous commit and adjust it based on https://github.com/php-imagine/Imagine/pull/843#discussion_r1143490195

ausi (approved):

PR #853, opened 2023-11-17
feat: adds NotFoundException when file does not exit

playmono:


mlocati:
For backward compatibility, I'd make the new `NotFoundException` exception extending `Imagine\Exception\InvalidArgumentException` and not `Imagine\Exception`

PR #855, opened 2023-12-02
Add PHP 8.3 to CI

dereuromark:
And it is sufficient to test each min/max of minors, even the min/max of majors usually suffices within PHP.

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -32,7 +32,7 @@ jobs:
       matrix:
         php-version:
           - "7.4"
-          - "8.1"
+          - "8.3"
         extensions:
           - gd,imagick
     steps:
@@ -87,12 +87,9 @@ jobs:
           - "5.5"
           - "5.6"
           - "7.0"
-          - "7.1"
-          - "7.2"
-          - "7.3"
           - "7.4"
           - "8.0"
-          - "8.1"
+          - "8.3"
         image-suffix:
           - gd-gmagick
           - gd-imagick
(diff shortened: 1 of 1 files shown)


mlocati:
> And it is sufficient to test each min/max of minors, even the min/max of majors usually suffices within PHP.

I'd disagree: it's better to test every single Major.Minor version of PHP.

dereuromark:
There is no logical reason in doing so, I can say that with confidence over thousands of repos and years of experience
Only creates more waiting time and burns resources :)
But do what you must.

dereuromark:
Your problem here seems to be of different kind anyway.
I will close this then and leave this to the maintainers.

mlocati:
> There is no logical reason in doing so, I can say that with confidence over thousands of repos and years of experience

I can say that 35 years of coding have taught me that it's better to never assume anything :wink:



> Your problem here seems to be of different kind anyway.

What do you mean?

dereuromark:
Well, I still have worked professionally for decades in that specific field :)
Assuming is the thing you do right now

> What do you mean?

See failing CI for 8.3

mlocati:
> Well, I still have worked professionally for decades in that specific field :)

Me too.

> Assuming is the thing you do right now

Yep, both for work (daily) and for a ton of open source projects.

I don't get the reason why you are so malign: you should reconsider the way you approach people you don't know.

dereuromark:
There is no malign. I am telling you the facts and logical reason.

I find it rather not helpful to have experts giving you advice and trying to help you keeping this (currently not so well maintained) repo maintained such an answer.
So count me out - good luck moving forward.

PR #858, opened 2024-02-17
Add support for webp_lossless to GD and GMagick

### LINKED ISSUE #857: Saving Lossless Webp?
### Issue description

Since php 8.1 the php GD library supports the saving of images in the webp lossless format using the construct:
```
imagewebp($image, 'image.webp', IMG_WEBP_LOSSLESS);
```
where IMG_WEBP_LOSSLESS is a constant set to the value 101.

The `saveOrOutput` method in the GD\Image.php class has an explicit check (line 734) to prevent the value of the quality parameter exceeding the bounds 0-100.

How in this case then can lossless webp images be saved?

ausi:
Fixes #857 

The IMagick driver already supports the `webp_lossless` option, see https://github.com/php-imagine/Imagine/blob/922e0327a6b36d1da0acca69993ac85395adcea6/src/Imagick/Image.php#L829-L831

This pull request adds support for GD and GMagick too.

### DIFF
--- a/src/Gd/Image.php
+++ b/src/Gd/Image.php
@@ -730,7 +730,9 @@ final class Image extends AbstractImage implements InfoProvider
                         $options['webp_quality'] = $options['quality'];
                     }
                 }
-                if (isset($options['webp_quality'])) {
+                if (!empty($options['webp_lossless'])) {
+                    $result[] = defined('IMG_WEBP_LOSSLESS') ? IMG_WEBP_LOSSLESS : 100;
+                } elseif (isset($options['webp_quality'])) {
                     if ($options['webp_quality'] < 0 || $options['webp_quality'] > 100) {
                         throw new InvalidArgumentException('webp_quality option should be an integer from 0 to 100');
                     }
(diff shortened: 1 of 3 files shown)


jcogs-design:
Actually this doesn't fix #857. Not sure why anyone would think it does. It fixes a related problem in a different library, but to suggest this is a fix for #857 is just wrong.

ausi:
> It fixes a related problem in a different library, but to suggest this is a fix for #857 is just wrong.

I don’t understand. In #857 you wrote about _“The `saveOrOutput` method in the GD\Image.php class has an explicit check (line 734)”_. And this pull request fixes exactly that as it adds support for the `webp_lossless` option on line 733.

What do you mean by _“different library”_? 
This pull request adds support for `webp_lossless` to GMagick **and** the GD library.

Did you test these changes?
The following code worked fine in my test:

```php
$image = (new \Imagine\Gd\Imagine)->open('image.png');
$image->save('image.webp', ['webp_lossless' => true]);
```

If it doesn’t work for you, please provide the steps to reproduce the issue.

PR #861, opened 2024-07-25
Use streams in ExifMetadataReader

### LINKED ISSUE #859: ExifMetadataReader class fails due to unnecessary check for 'allow_url_fopen()'
### Issue description
A new Imagine instance can be generated by loading the content of the image file directly.
If you try and add ExifMetadataReader to an instance created this way, and the server environment has `allow_url_fopen` set to 0, the attempt fails: even though in this instance there is no need for remote access to anything to generate the image or read the metadata.

The issue is being felt on a server that has a 'locked down' security status (i.e. allow_url_fopen is set to 0, but curl is available) - even though in the specific use case no remote file access is required, if it was Curl is available and could be used. But the check carried out by ExifMetadataReader in `getUnsupportedReason()` doesn't consider either of these contexts, and throws the exception simply because of the value of `allow_url_fopen` which appears to be inappropriate.
...

### What version of Imagine are you using?

1.3.5

### What's the PHP version you are using?

8.3

### What's the imaging library you are using [gd/imagick/gmagick/any]?

gd2

### Minimal PHP code to reproduce the error: 

Where `$source_image_raw` is the content of an image file (obtained from reading a local disk file, or from remote location).

```php

$source_image = (new Imagine())->setMetadataReader(new \Imagine\Image\Metadata\ExifMetadataReader())->load($source_image_raw);
```

ausi:
Fix #859
Alternative to #860

### DIFF
--- a/src/Image/Metadata/ExifMetadataReader.php
+++ b/src/Image/Metadata/ExifMetadataReader.php
@@ -42,11 +42,14 @@ class ExifMetadataReader extends AbstractMetadataReader
         if (!function_exists('exif_read_data')) {
             return 'The PHP EXIF extension is required to use the ExifMetadataReader';
         }
-        if (!in_array('data', stream_get_wrappers(), true)) {
-            return 'The data:// stream wrapper must be enabled';
-        }
-        if (in_array(ini_get('allow_url_fopen'), array('', '0', 0), true)) {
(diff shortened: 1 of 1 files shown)


ausi:
Failing CI seems to be unrelated

PR #864, opened 2024-11-29
[FEATURE] Run tests against PHP > 8.1

NicoHaase:


mlocati:
I think we also need to update https://github.com/php-imagine/docker-builder

NicoHaase:
> I think we also need to update [php-imagine/docker-builder](https://github.com/php-imagine/docker-builder)

Yep, looks like that... but I don't really get how to fix the workflow (see https://github.com/NicoHaase/docker-builder/actions) - a lot of the older PHP versions won't built properly

mlocati:
> Yep, looks like that... but I don't really get how to fix the workflow (see https://github.com/NicoHaase/docker-builder/actions) - a lot of the older PHP versions won't built properly

Fixed

mlocati:
Superseded by #865

PR #865, opened 2024-12-02
Fix PHP 8.4 deprecations nullable objects

alexander-schranz:
This should avoid outpus like:

> Deprecated: Imagine\Image\ImagineInterface::create(): Implicitly marking parameter $color as nullable is deprecated, the explicit nullable type must be used instead in /sulu-develop.localhost/vendor/imagine/imagine/src/Image/ImagineInterface.php on line 36
PHP Deprecated:  Imagine\Image\ManipulatorInterface::rotate(): Implicitly marking parameter $background as nullable is deprecated, the explicit nullable type must be used instead in /sulu-develop.localhost/vendor/imagine/imagine/src/Image/ManipulatorInterface.php on line 97

> Deprecated: Imagine\Image\ManipulatorInterface::rotate(): Implicitly marking parameter $background as nullable is deprecated, the explicit nullable type must be used instead in /sulu-develop.localhost/vendor/imagine/imagine/src/Image/ManipulatorInterface.php on line 97
PHP Deprecated:  FOS\RestBundle\Serializer\Normalizer\FormErrorHandler::serializeFormToXml(): Implicitly marking parameter $context as nullable is deprecated, the explicit nullable type must be used instead in /sulu-develop.localhost/vendor/friendsofsymfony/rest-bundle/Serializer/Normalizer/FormErrorHandler.php on line 42


ausi:
Nullable types are only supported since PHP 7.1.0 so we would need to increase the version constraint in https://github.com/php-imagine/Imagine/blob/8908a6cee4242a5fd4e4457ab5713449b1934cfa/composer.json#L20 I think.

alexander-schranz:
@mlocati should we increase min version?

mlocati:
I think so. PHP 5 is very very outdated (but I'd keep compatibility with PHP 7)

alexander-schranz:
@mlocati adopted. Think the rest of the errors are unrelated and in some cases connected to: https://github.com/php-imagine/docker-builder/pull/5

ausi (approved):

mlocati:
> in some cases connected to: [php-imagine/docker-builder#5](https://github.com/php-imagine/docker-builder/pull/5)

I've just published all the docker images required for tests

tests/tests/Gmagick/LayersTest.php
```diff
@@ -86,9 +86,11 @@ public function testCount()
 
     public function testWebpFormatIsAllowedAsAnimatedFormat()
     {
-        $palette = new RGB();
+        $this->checkGmagickMockable();
```
alexander-schranz, line 89: looks this was missed in another PR to be checked.

alexander-schranz:
Can somebody help me with:

```
1) Imagine\Test\Imagick\LayersTest::testWebpFormatIsAllowedAsAnimatedFormat
TypeError: Argument 1 passed to Imagine\Imagick\Image::__construct() must be an instance of Imagick, null given, called in /__w/Imagine/Imagine/src/Factory/ClassFactory.php on line 126

/__w/Imagine/Imagine/src/Imagick/Image.php:70
/__w/Imagine/Imagine/src/Factory/ClassFactory.php:126
/__w/Imagine/Imagine/src/Imagick/Layers.php:180
/__w/Imagine/Imagine/src/Imagick/Layers.php:163
/__w/Imagine/Imagine/src/Imagick/Layers.php:108
/__w/Imagine/Imagine/tests/tests/Imagick/LayersTest.php:97
phpvfscomposer:///__w/Imagine/Imagine/vendor/phpunit/phpunit/phpunit:97
```

mlocati:
> TypeError: Argument 1 passed to Imagine\Imagick\Image::__construct() must be an instance of Imagick, null given, called in /__w/Imagine/Imagine/src/Factory/ClassFactory.php on line 126

I think the simplest solution is to avoid mocks.

Something like this:

```php
    public function testWebpFormatIsAllowedAsAnimatedFormat()
    {
        $image = $this->getImagine()->open(IMAGINE_TEST_FIXTURESFOLDER . '/anima3.gif');
        $originalDelayInTicks = $image->layers()->get(0)->getImagick()->getImageDelay();
        $image->layers()->animate('webp', (int) (1000 + $originalDelayInTicks * 1000 / 20), 0);
        $this->assertSame('webp', $image->getImagick()->getFormat());
        $newDelayInTicks = $image->layers()->get(0)->getImagick()->getImageDelay();
        $this->assertNotEquals($originalDelayInTicks, $newDelayInTicks);
    }
```

ausi:
My guess would be that `->getImage()` returns `null` here https://github.com/php-imagine/Imagine/blob/8908a6cee4242a5fd4e4457ab5713449b1934cfa/src/Imagick/Layers.php#L180 because the mock object does not know about the correct return type?

Maybe adding the following lines to the test might solve it:

```php
$resource->expects($this->atLeastOnce())
    ->method('getImage')
    ->will($this->returnValue($resource));
```

alexander-schranz:
Thank you both. I removed did go with the mockless solution looks the CI is now green :tada:

mlocati:
Thank you!

PR #875, opened 2026-01-05
do not call curl_close on PHP 8.0+

### LINKED ISSUE #874: Fix php 8.5 deprecation about `curl_close`
I see this deprecation in my testsuite:

```
1) /var/www/app/vendor/imagine/imagine/src/File/Loader.php:243
Function curl_close() is deprecated since 8.5, as it has no effect since PHP 8.0
``` 

Since we are still supporting php 7 here we probably need to conditionally omit the call?

dmaicher:
Fixes https://github.com/php-imagine/Imagine/issues/874


ausi (approved):

mlocati:
Thanks!

PR #877, opened 2026-06-03
Preserve alpha channel in Imagick negative effect

nlemoine:
`effects()->negative()` on the Imagick driver inverts the alpha channel as well as the colours, which GD doesn't do. Result: an opaque image comes back fully transparent. The colour values are right, you just can't see them.

Cause is `negateImage(false, \Imagick::CHANNEL_ALL)`. `CHANNEL_ALL` includes alpha, so opaque alpha flips to transparent. GD avoids this because `IMG_FILTER_NEGATE` only touches colour.

Fix is to negate with `CHANNEL_ALL & ~CHANNEL_ALPHA` so only the colour channels flip (same thing intervention/image does). The Gmagick driver doesn't need it: in the gd-gmagick docker image `CHANNEL_ALL` leaves an opaque image opaque, and Gmagick's CHANNEL constants are sequential enum values rather than bitmasks, so the `& ~` trick wouldn't apply anyway.

Added `testNegatePreservesAlpha` to `AbstractEffectsTest` so every driver gets covered. It negates a fully-opaque image and asserts it stays opaque. Fails on Imagick before the fix (alpha 100 to 0), passes after. Verified the full suite green in both the gd-imagick and gd-gmagick docker images.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ### NEXT (YYYY-MM-DD)
+- Fix the Imagick driver inverting the alpha channel in `effects()->negative()` (an opaque image became fully transparent); it now excludes the alpha channel, matching the GD driver
 
 ### 1.5.2 (2026-01-09)
 - Do not call curl_close on PHP 8.0+ (#875, @dmaicher)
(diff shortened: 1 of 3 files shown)


mlocati:
Thanks!

nlemoine:
Thanks for merging this so fast @mlocati!

PR #878, opened 2026-06-03
Preserve transparency when pasting with alpha on the Imagick driver

nlemoine:
`paste()` with an alpha < 100 on the Imagick driver paints a "black box" over the transparent areas of the pasted image. Paste a watermark that has a  transparent surround at, say, 50% opacity and the whole bounding box comes back as a semi-opaque dark rectangle instead of just the visible part fading.

Culprit is `setImageAlpha($alpha / 100)` (and the `setImageOpacity` fallback). Both set every pixel's alpha to the same value, so fully-transparent pixels turn semi-opaque. GD doesn't have this problem because `imagecopymerge` blends without flattening the source alpha.

The fix scales the existing alpha instead of overwriting it: `evaluateImage(EVALUATE_MULTIPLY, $alpha / 100, CHANNEL_ALPHA)`, after activating the alpha channel. Transparent stays transparent, opaque fades.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ### NEXT (YYYY-MM-DD)
+- Fix the Imagick driver painting a "black box" when pasting an image with transparent areas at an alpha lower than 100; the opacity now scales the existing per-pixel alpha instead of overwriting it
 
 ### 1.5.2 (2026-01-09)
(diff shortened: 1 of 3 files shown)


mlocati:
👍

nlemoine:
Thank you @mlocati!

mlocati:
Thanks to you for these fixes and related tests!

PR #879, opened 2026-06-04
Fix imagick stale layers

nlemoine:
Resize an animated GIF on the Imagick driver, then ask for its layers: you get the pre-resize frames back.

```php
$image = $imagine->open('animated.gif'); // 2 frames, 8x8
$image->resize(new Box(4, 4));

$image->getSize()->getWidth();                                // 4
$image->layers()->coalesce()->get(0)->getSize()->getWidth();  // 8, expected 4
```

`Layers` grabs the `\Imagick` instance in its constructor, and `Image::layers()` caches the object forever. That was harmless until the multi-frame paths started swapping the instance: `coalesceImages()`/`deconstructImages()` return a new `\Imagick`, which `resize()`, `crop()`, `flatten()` and the `optimize` save path assign to `$this->imagick`, so the cached `Layers` keeps reading frames nothing else uses anymore. One concrete casualty: `animated.delay`/`animated.loops` are applied through `layers()` in `prepareOutput()`, land on the orphaned frames, and the saved file keeps its original timing.

Reassignments now go through a small setter that drops the cached `Layers` when the instance changes (`__clone()` already rebuilds it the same way), so the next `layers()` call reads the live resource.

### DIFF
--- a/src/Imagick/Image.php
+++ b/src/Imagick/Image.php
@@ -126,6 +126,20 @@ final class Image extends AbstractImage implements InfoProvider
     }
 
     /**
+     * Replaces the underlying \Imagick instance, invalidating the cached
+     * layers (if any) since they wrap the replaced instance.
+     *
+     * @param \Imagick $imagick
+     */
+    private function setImagick(\Imagick $imagick)
+    {
+        if ($imagick !== $this->imagick) {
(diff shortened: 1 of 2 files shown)


mlocati:
Nice catch!

PR #880, opened 2026-06-04
Preserve transparency in the Imagick grayscale effect

nlemoine:
`$image->effects()->grayscale()` on the Imagick driver turns transparent pixels into opaque gray. `Imagick::IMGTYPE_GRAYSCALE` is one of ImageMagick's alpha-less image types, and switching to it deactivates the image's alpha channel (`getImageAlphaChannel()` flips from `true` to `false`). The pixels keep their stored alpha values in memory — only the encoders ignore them — so the breakage only surfaces in the saved file, whatever the format. That's also why the test asserts through an encode/decode roundtrip instead of reading pixels in place.

The driver already solves this exact problem in `Image::setColorspace()`, where `usePalette()` with a grayscale palette deliberately lands on `IMGTYPE_GRAYSCALEALPHA` to keep alpha; `effects()->grayscale()` just never got the same treatment. It now picks its type through the same fallback chain (`IMGTYPE_GRAYSCALEALPHA` on ImageMagick 7 / Imagick 3.4.3+, `IMGTYPE_GRAYSCALEMATTE` before, hard-coded `3` when neither constant exists). The gray rendition is unchanged — both types go through the same colorspace transform, only the alpha trait differs — so the existing `testGrayscale` expectations still hold.

GD is unaffected (`IMG_FILTER_GRAYSCALE` preserves alpha), and Gmagick doesn't support transparency at all, so the shared test skips there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ### NEXT (YYYY-MM-DD)
+- Fix the Imagick driver deactivating the alpha channel in `effects()->grayscale()` (transparent pixels were encoded as opaque gray); it now uses the alpha-preserving grayscale image type, like `usePalette()` (#880, @nlemoine)
 
 ### 1.5.3 (2026-06-03)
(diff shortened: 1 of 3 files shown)


mlocati:
Thanks!

nlemoine:
Thanks @mlocati :)

